PR #30, opened 2016-06-29
Add set scheme client option

aanm:
Signed-off-by: André Martins <aanm90@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -88,6 +88,7 @@ type Client struct {
 	setContentLength bool
 	isHTTPMode       bool
 	outputDirectory  string
+	scheme           string
 	proxyURL         *url.URL
 	beforeRequest    []func(*Client, *Request) error
 	afterResponse    []func(*Client, *Response) error
@@ -582,6 +583,17 @@ func (c *Client) SetTransport(transport *http.Transport) *Client {
 	return c
 }
 
+// SetScheme method sets custom scheme in the resty client. Its way to override default.
+// 		resty.SetScheme("http")
+//
+func (c *Client) SetScheme(scheme string) *Client {
+	if c.scheme == "" {
+		c.scheme = scheme
+	}
+
+	return c
+}
+
 // executes the given `Request` object and returns response
 func (c *Client) execute(req *Request) (*Response, error) {
 	// Apply Request middleware
@@ -603,6 +615,7 @@ func (c *Client) execute(req *Request) (*Response, error) {
 
 	req.Time = time.Now()
 	c.httpClient.Transport = c.transport
+
 	resp, err := c.httpClient.Do(req.RawRequest)
 
 	response := &Response{
--- a/resty_test.go
+++ b/resty_test.go
@@ -1213,6 +1213,14 @@ func TestSetTransport(t *testing.T) {
 	assertEqual(t, true, DefaultClient.transport != nil)
 }
 
+func TestSetScheme(t *testing.T) {
+	DefaultClient = dc()
+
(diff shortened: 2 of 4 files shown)


client.go
```diff
@@ -603,6 +615,12 @@ func (c *Client) execute(req *Request) (*Response, error) {
 
 	req.Time = time.Now()
 	c.httpClient.Transport = c.transport
+
+	if req.RawRequest.URL != nil && req.RawRequest.URL.Scheme == "" {
```
jeevatkm: Can you please move this complete `if` condition to [middleware.go#L141](https://github.com/go-resty/resty/blob/master/middleware.go#L141)?

client.go
```diff
@@ -603,6 +615,12 @@ func (c *Client) execute(req *Request) (*Response, error) {
...
+		req.RawRequest.URL.Scheme = c.scheme
+		req.RawRequest.URL.Host = req.URL
```
jeevatkm: why this line is needed, can you help me to understand?

jeevatkm:
@aanm Thank you so much for contributing to `Resty`.

I have added couple of note, can you please have a look and take care? Also test case please for your scenario.

client.go
aanm: An `URL` is not valid with `scheme` only, the host is also needed

aanm:
@jeevatkm Done

client.go
jeevatkm: Thanks, so you need only `Host` here correct?

If yes `req.URL` has complete request URL, you may need to parse `Host` from `req.URL`. Please correct me if I missing something.

jeevatkm:
@aanm Thanks for your effort, I appreciate it. Travis CI check didn't pass, added a not for URL and test case.

resty_test.go
```diff
@@ -1213,6 +1213,16 @@ func TestSetTransport(t *testing.T) {
...
 }
 
+func TestSetScheme(t *testing.T) {
+	ts := createGetServer(t)
+	defer ts.Close()
+	DefaultClient = dc()
+
+	SetScheme("http")
+
+	assertEqual(t, true, DefaultClient.scheme == "true")
```
jeevatkm: this line may be -

`assertEqual(t, true, DefaultClient.scheme == "http")`

resty_test.go
```diff
@@ -1213,6 +1213,16 @@ func TestSetTransport(t *testing.T) {
 	assertEqual(t, true, DefaultClient.transport != nil)
 }
 
+func TestSetScheme(t *testing.T) {
+	ts := createGetServer(t)
```
jeevatkm: line # 1216 and # 1217 may not be needed for this test case.

client.go
aanm: @jeevatkm Not for my use case.

The requested URL won't contain any `Host`, that's why I had to set up the `Scheme` like that since the transport is given by me.

This allows go-resty to be used with unix sockets.

aanm:
@jeevatkm Fixed

jeevatkm:
@aanm Thanks, this is helpful for community.

> This allows go-resty to be used with unix sockets.

If you're interested, please add example too for community. I'm glad to this PR and your upcoming examples.

jeevatkm:
@aanm Would you be interested in providing example for `resty with unix socket`?

Like creating an `example` directory and within that `resty_unix_socket.go`

aanm:
@jeevatkm https://github.com/go-resty/resty/pull/32

PR #31, opened 2016-07-07
Option to close connection for each HTTP request

aanm:
Signed-off-by: André Martins <aanm90@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -91,6 +91,7 @@ type Client struct {
 	scheme           string
 	proxyURL         *url.URL
 	mutex            *sync.Mutex
+	closeConnection  bool
 	beforeRequest    []func(*Client, *Request) error
 	afterResponse    []func(*Client, *Response) error
 }
@@ -592,6 +593,13 @@ func (c *Client) SetScheme(scheme string) *Client {
 		c.scheme = scheme
 	}
 
+	return c
+}
+
+// SetCloseConnection method sets variable Close in http request struct with the given
+// value. More info: https://golang.org/src/net/http/request.go
+func (c *Client) SetCloseConnection(close bool) *Client {
+	c.closeConnection = close
 	return c
 }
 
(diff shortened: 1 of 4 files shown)


jeevatkm:
@aanm Thanks for the PR.

Technically there is no leak as per golang core.  Below note taken from golang godoc-
```go
// For client requests, setting this field prevents re-use of
// TCP connections between requests to the same hosts, as if
// Transport.DisableKeepAlives were set.
Close bool
```

So there is no issue with `go-resty` library. I will be renaming the PR subject.

----

I would say it up to developer or user choose the behavior they need. My recommendation is to expose some sort setter method to `opt-in` if they want to close connection for each request and defaults to false (also same as golang core).

your thoughts?

aanm:
@jeevatkm True, on my use case all connection were for the same host but with multiple resty clients. Setting that option to true dropped the number of connections open by 70%. I'll think on something an make it as an option. Any ideas where it would be the best place to put it?

jeevatkm:
You have specific use case. No issues. Best place is to -

* Create a `field` called `closeConnection` after `mutex` [field](https://github.com/go-resty/resty/blob/master/client.go#L93)
* Add `SetCloseConnection` for client struct like 
   * `func (c *Client) SetCloseConnection(close bool) *Client`
   * Please add nice godoc documentation, so user can understand the purpose
* Initialize explicitly on client creation after `mutex` [field](https://github.com/go-resty/resty/blob/master/default.go#L39) like `closeConnection: false`
* Also please add `SetCloseConnection` method to `default.go` like you did in previous PR

finally assign in it the request (you already pick spot for this PR, but small change):
```go
r.RawRequest.Close = c.closeConnection
```

jeevatkm:
Test case please and watch out travis ci build.

aanm:
@jeevatkm ping

Issue #44, opened 2016-11-25
Context support


ei-grad:
It would be nice to have a Request.SetContext method to set context for request cancellation like it is done in httpDo function described in https://blog.golang.org/context.

ei-grad:
Oops, it would break golang <1.7, is it planned to keep support for them?

jeevatkm:
Support provide in PR #46. Merged to master.

PR #46, opened 2016-11-25
Add Request.SetContext

ei-grad:
Implements #44

### DIFF
--- a/context17_test.go
+++ b/context17_test.go
@@ -1,0 +1,9 @@
+// +build !go1.8
+
+package resty
+
+import "strings"
+
+func errIsContextCanceled(err error) bool {
+	return strings.Contains(err.Error(), "request canceled")
+}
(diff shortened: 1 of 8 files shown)


jeevatkm:
@ei-grad Thanks for your contribution. 

Yes this feature only supported Go 1.7 and above. So we may have to limit this option. Let me have a look.

jeevatkm:
@ei-grad I have quick question for this PR.

To achieve this PR to work properly, following needs to be done:
* Please branch out fresh, since code quality improvements applied in `master` brach
* We have to take advantage of Go build constraints [reference](https://golang.org/pkg/go/build/)
* Create two files for `type Request struct` with definition for `go1.7 & greater`, `go1.6 & below` and implementation will be separate file which is common part
* And test cases accordingly

This approach is compatible with all Go versions.

Would you like like to try and send PR?

ei-grad:
Yep, I'll do that :).

ei-grad:
@jeevatkm done

PR #51, opened 2017-01-05
Request based on the SRV record of the service

DennyLoko:
I've made this changes to be possible to do a request based on the SRV record of the service. It is useful when you have a service discovery mechanism, like Consul, so you can pass the information and the system will query the SRV record and to the request based on it.

I do not expect this PR to be merged right away, but to start a discussion on the following topic.

Please, review it and suggest the needed changes.

### DIFF
--- a/request.go
+++ b/request.go
@@ -11,10 +11,17 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"net"
 	"net/url"
 	"reflect"
 	"strings"
 )
+
+// SRVRecord holds the data to query the SRV record for the following service
+type SRVRecord struct {
+	Service string
+	Domain  string
+}
 
 // SetHeader method is to set a single header field and its value in the current request.
 // Example: To set `Content-Type` and `Accept` as `application/json`.
@@ -345,6 +352,16 @@ func (r *Request) SetProxy(proxyURL string) *Request {
 	return r
 }
 
+// SetSRV method sets the details to query the service SRV record and execute the
+// request.
+// 		resty.R().
+//			SetSRV(SRVRecord{"web", "testservice.com"}).
+//			Get("/get")
+func (r *Request) SetSRV(srv *SRVRecord) *Request {
+	r.SRV = srv
+	return r
+}
+
 //
 // HTTP verb method starts here
 //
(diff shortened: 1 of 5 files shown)


DennyLoko:
One of the things that I did not implement is the round robin based on the weight of the record. 

The [`net.LookupSRV`](https://golang.org/pkg/net/#LookupSRV) function already returns the records sorted by priority and randomize the ones with the same weight. So, upon the retry, it will try another service based on this criteria.

jeevatkm:
@DennyLoko Thanks for the PR, I appreciate it.

I know `SRV` records lookup is widely used in XMPP/Jabber, SIP, LDAP, etc.; not sure about HTTP(s) world.

Can you please help me to understand, this use case is benefits widely?

DennyLoko:
Hello @jeevatkm!

We use an extensive number of Docker containers on our microservices infrastructure, and we use [Consul](https://www.consul.io) as service discovery mechanism to help us reach each container without the need to have a list of IP addresses of each container of each service.

Consul provides and DNS interface for querying the `SRV` record of the services, so this change could help other users user which has the same infrastructure as me.

jeevatkm:
Thanks for providing real usage scenario. I hope and believe it will be useful for `resty` users. I will go through PR code changes and share my suggestions.

Thanks again for your time and effort.

request.go
```diff
@@ -465,3 +489,15 @@ func (r *Request) fmtBodyString() (body string) {
...
+func (r *Request) selectAddr(addrs []*net.SRV, path string, attempt int) string {
+	if addrs == nil {
+		return path
+	}
+
+	idx := attempt % len(addrs)
+	domain := strings.TrimRight(addrs[idx].Target, ".")
+	path = strings.TrimLeft(path, "/")
+
+	return fmt.Sprintf("%s://%s:%d/%s", DefaultClient.scheme, domain, addrs[idx].Port, path)
```
jeevatkm: Please use `r.client.scheme` instead of `DefaultClient.scheme`.

request.go
```diff
@@ -465,3 +489,15 @@ func (r *Request) fmtBodyString() (body string) {
 
 	return
 }
+
+func (r *Request) selectAddr(addrs []*net.SRV, path string, attempt int) string {
+	if addrs == nil {
```
jeevatkm: Can you please add test case for when `addrs` is `nil` and SRV is record is set. So we can ensure request goes through or gives error info to caller.

jeevatkm:
@DennyLoko whenever you push your changes, I will merge it. Thanks.

request.go
DennyLoko: Maybe I didn't understand, but I think [this][1] test case already cover this situation. 

[1]: https://github.com/go-resty/resty/pull/51/files/9a10560ae3e2c992116bf965323d403082b7ad5c#diff-283b9a8bcf7d3d14aaaa296a2bfd884cR1428

jeevatkm:
@DennyLoko no worries, thanks for your response.

Changes looks good, can you please squash these commits?

DennyLoko:
Done 👍

jeevatkm:
@DennyLoko Thanks, PR is merged.

Issue #59, opened 2017-03-10
Lost repsonse and no error return


lrita:
Someone api will cost N ms.

If we `SetTimeout` M ms (N< M < 2N) to `Client` and we call this api twice,  the server side will receive three request and send three response. 

In client side, user only see 2 api call but 3 api call be sent actually.

This issue will be duplicated in go 1.7.3, [here is the code](https://github.com/lrita/resty/blob/fix-timeout-issue/client_test.go#L101-L120).

I found a bad implement of `SetTimeout` and fix it.

jeevatkm:
@lrita Thanks for your PR, I have added a comment. Please have a look.

lrita:
I have fixed the build errors in golang 1.2.

jeevatkm:
Thanks @lrita.

jeevatkm:
PR #60 Merged.

PR #60, opened 2017-03-10
Use HTTP Client SetTimeout instead of custom timeout

lrita:
fix(#59) by using `http.Client.Timeout` instead of a special `http.Transport`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -491,14 +491,6 @@ func (c *Client) SetTLSClientConfig(config *tls.Config) *Client {
 	return c
 }
 
-// SetTimeout method sets timeout for request raised from client
-//		resty.SetTimeout(time.Duration(1 * time.Minute))
-//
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -496,17 +495,7 @@ func (c *Client) SetTLSClientConfig(config *tls.Config) *Client {
 //		resty.SetTimeout(time.Duration(1 * time.Minute))
 //
 func (c *Client) SetTimeout(timeout time.Duration) *Client {
```
jeevatkm: Currently `resty` support Go1.2 and above. HTTP Client `SetTimeout` was introduced in Go1.3. You see [travis build is failed for Go1.2](https://travis-ci.org/go-resty/resty/jobs/209669843)

Can you please handle these support with build constraints?

PR #66, opened 2017-04-07
Allow to override default retry wait times for Client

### LINKED ISSUE #64: Add support for WaitTime and MaxWaitTime for retries
Currently client does not expose a way to configure wait times for retry, only conditions and retry count are configurable - https://github.com/go-resty/resty/blob/90b37fd783f8bd880cf2bd537a2c3b3b48cf32e1/request.go#L430

We need to make wait times configurable as well, default 100ms wait time is too little in my case.

I can provide a pull request for this if the issue is valid

bak1an:
This resolves #64 

Changes made here:

- Changing retry wait time options to have `time.Duration` type instead of of `int` holding milliseconds.

- Adding `RetryWaitTime` and `RetryMaxWaitTime` fields to `Client` along with chaining setters.

- A condition to make sure `Backoff` always sleeps at least `WaitTime` duration. Before this change it could sleep less than this due to jitter we have there. Especially before the very first retry, the sleep duration was always somewhere between `WaitTime / 2` and `WaitTime`.

- Test to make sure overridden wait times are actually used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ resty tested with Go `v1.2` and above.
     * FlexibleRedirectPolicy
     * DomainCheckRedirectPolicy
     * etc. [more info](redirect.go)
-  * Retry Mechanism [how to use](retry_test.go)
+  * Retry Mechanism [how to use](#retries)
     * Backoff Retry
     * Conditional Retry
   * SRV Record based request instead of Host URL [how to use](resty_test.go#L1412)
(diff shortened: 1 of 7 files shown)


bak1an:
@jeevatkm Makes sense.

I'll add some info on retries to readme.

jeevatkm:
Thank you @bak1an.

jeevatkm:
@bak1an Shall I go ahead and merge this PR or wait for your readme changes? Please let me know.

bak1an:
@jeevatkm 

Sorry for the delay,

Added some retry usage doc and mentioned delay defaults in doc.

jeevatkm:
Thank you @bak1an, I appreciate it.

PR #69, opened 2017-04-28
add SetCookieJar method

lrita:
* allow user override default CookieJar in resty client

### DIFF
--- a/default.go
+++ b/default.go
@@ -89,6 +89,11 @@ func SetHeaders(headers map[string]string) *Client {
 	return DefaultClient.SetHeaders(headers)
 }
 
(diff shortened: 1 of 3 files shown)


default.go
```diff
@@ -89,6 +89,11 @@ func SetHeaders(headers map[string]string) *Client {
 	return DefaultClient.SetHeaders(headers)
 }
 
+// SetCookieJa sets custom http.CookieJar. See `Client.SetCookieJar` for more information.
```
jeevatkm: Just one typo error in godoc `SetCookieJa` => `SetCookieJar`.

PR #70, opened 2017-04-29
WIP Add url parameter

abihf:
Example: `resty.R().SetParam("id", "123").Get("/users/:id")`
Will become: `/users/123`

### DIFF
--- a/client.go
+++ b/client.go
@@ -282,6 +282,7 @@ func (c *Client) R() *Request {
 	r := &Request{
 		URL:            "",
 		Method:         "",
+		Param:          make(map[string]string),
 		QueryParam:     url.Values{},
 		FormData:       url.Values{},
 		Header:         http.Header{},
--- a/middleware.go
+++ b/middleware.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"strings"
 )
 
@@ -43,6 +44,15 @@ func parseRequestURL(c *Client, r *Request) error {
 			return err
 		}
 	}
+
+	path := reqURL.Path
+	for k, v := range r.Param {
+		re, err := regexp.Compile(`:` + k + `\b`)
+		if err == nil {
+			path = re.ReplaceAllString(path, v)
+		}
+	}
+	reqURL.Path = path
 
 	// Adding Query Param
 	query := reqURL.Query()
--- a/request.go
+++ b/request.go
@@ -54,6 +54,32 @@ func (r *Request) SetHeaders(headers map[string]string) *Request {
 	return r
 }
 
+// SetParam method sets single url paramater and its value in the current request.
+// It will be formed as query string for the request.
+// Example: `/user/:id` will become `/user/123`
+// 		resty.R().
+//			SetParam("id", "123").
+// Also you can override query params value, which was set at client instance level
+//
+func (r *Request) SetParam(name, value string) *Request {
+	r.Param[name] = value
+	return r
+}
+
+// SetParams method sets multiple url paramaters and its value in the current request.
+// It will be formed as query string for the request.
+// Example: `/user/:id/posts/:postId` will become `/user/123/post/321`
+// 		resty.R().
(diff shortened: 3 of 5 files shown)


middleware.go
```diff
@@ -44,6 +45,15 @@ func parseRequestURL(c *Client, r *Request) error {
 		}
 	}
 
+	path := reqURL.Path
+	for k, v := range r.Param {
```
jeevatkm: Let's call this `:id` as `PathParam` and `args` as for those values.

* It's easily error prone for users, so validation is required/improtant
     * PathParam count and args count (refer codesnippet: `countPathParams`) should match otherwise return error
     * Keep sane number, let's limit path param count to `100`, if path params is more than 100 return error
     * if path param is missing value then return error
* Regex is not needed here. Also it's expensive operation
     * refer codesnippet: `composePathURL`

**Please note:** I have not tested these code snippet, but it expected to work. if needs correction please feel free to do so.

**Code Snippet:** `countPathParams`
```go
func countPathParams(path string) uint8 {
  var n uint
  for i := 0; i < len(path); i++ {
  	if path[i] != ':' {
  		continue
  	}
  	n++
  }
  return uint8(n)
}
```

**Code Snippet:** `composePathURL`
```go
// compose path URL with values
pathURL := "/"
for _, segment := range strings.Split(reqURL.Path, "/")[1:] {
  if len(strings.TrimSpace(segment)) == 0 {
    continue
  }

  if segment[0] == ':' {
    argName := segment[1:]
    if arg, found := r.PathParam[argName]; found {
      pathURL = path.Join(pathURL, fmt.Sprintf("%v", arg))
      delete(r.PathParam, argName)
      continue
    }

    return fmt.Errorf("'%v' param not found", segment[1:])
  }

  pathURL = path.Join(pathURL, segment)
}

reqURL.Path = pathURL
```

client.go
```diff
@@ -282,6 +282,7 @@ func (c *Client) R() *Request {
 	r := &Request{
 		URL:            "",
 		Method:         "",
+		Param:          make(map[string]string),
```
jeevatkm: Let's keep variable as `PathParam`

request.go
```diff
@@ -54,6 +54,32 @@ func (r *Request) SetHeaders(headers map[string]string) *Request {
...
 }
 
+// SetParam method sets single url paramater and its value in the current request.
+// It will be formed as query string for the request.
+// Example: `/user/:id` will become `/user/123`
+// 		resty.R().
+//			SetParam("id", "123").
+// Also you can override query params value, which was set at client instance level
+//
+func (r *Request) SetParam(name, value string) *Request {
```
jeevatkm: Let's keep method name as `SetPathParam`

request.go
```diff
@@ -54,6 +54,32 @@ func (r *Request) SetHeaders(headers map[string]string) *Request {
...
+}
+
+// SetParams method sets multiple url paramaters and its value in the current request.
+// It will be formed as query string for the request.
+// Example: `/user/:id/posts/:postId` will become `/user/123/post/321`
+// 		resty.R().
+//			SetParams(map[string]string{"id":"123", "postId": "321"}).
+// Also you can override query params value, which was set at client instance level
+//
+func (r *Request) SetParams(params map[string]string) *Request {
```
jeevatkm: Let's keep method name as `SetPathParams`

request16.go
```diff
@@ -23,6 +23,7 @@ import (
 type Request struct {
 	URL        string
 	Method     string
+	Param      map[string]string
```
jeevatkm: Let's keep variable name as `PathParam`

request17.go
```diff
@@ -24,6 +24,7 @@ import (
 type Request struct {
 	URL        string
 	Method     string
+	Param      map[string]string
```
jeevatkm: Let's keep variable name as `PathParam`

jeevatkm:
@abihf Looking forward to your response.

jeevatkm:
@abihf Looking forward to your response.

jeevatkm:
Closing due to non-response. Thanks.

PR #71, opened 2017-05-01
Rewire set methods to actual url.Values' set methods

alexmontecucco:
Rewire set methods to actual url.Values' set methods for:
 - setFormData
 - setQueryParams

 + a few typo fix

### DIFF
--- a/client.go
+++ b/client.go
@@ -161,7 +161,7 @@ func (c *Client) SetCookieJar(jar http.CookieJar) *Client {
 	return c
 }
 
-// SetCookie method sets a single cookie in the client instance.
+// SetCookie method appends a single cookie in the client instance.
 // These cookies will be added to all the request raised from this client instance.
 // 		resty.SetCookie(&http.Cookie{
 // 					Name:"go-resty",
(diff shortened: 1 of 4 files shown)


jeevatkm:
@alexmontecucco Thank you for the PR.

client.go
```diff
@@ -236,13 +236,13 @@ func (c *Client) SetQueryParam(param, value string) *Client {
 //
 func (c *Client) SetQueryParams(params map[string]string) *Client {
 	for p, v := range params {
-		c.QueryParam.Add(p, v)
+		c.QueryParam.Set(p, v)
```
jeevatkm: Shall we do, as same as `Request` object method?

```go
c.SetQueryParam(p, v)
```

request.go
```diff
@@ -48,13 +48,13 @@ func (r *Request) SetHeader(header, value string) *Request {
 //
 func (r *Request) SetHeaders(headers map[string]string) *Request {
 	for h, v := range headers {
-		r.Header.Set(h, v)
+		r.SetHeader(h, v)
```
jeevatkm: This is good. Thanks.

request.go
```diff
@@ -79,13 +79,13 @@ func (r *Request) SetQueryParam(param, value string) *Request {
 //
 func (r *Request) SetQueryParams(params map[string]string) *Request {
 	for p, v := range params {
-		r.QueryParam.Add(p, v)
+		r.SetQueryParam(p, v)
```
jeevatkm: This is good, thanks.

client.go
alexmontecucco: Yes, missed, will do

PR #73, opened 2017-05-28
Add shortcuts to set DefaultClient's retry timeouts

bak1an:
Just noticed we have these package level shortcuts for configuring DefaultClient and added them for retry timeouts.


jeevatkm:
@bak1an PR looks good. Just we need to cover test coverage.

PR #77, opened 2017-06-16
Why set a UA if one is already supplied

Robbilie:
some monitoring software has to prioritize User-Agent, X-User-Agent or possibly some query string parameter where youd typically rank qs and X-User-Agent a higher prio than the normal UA because in some environments (browsers) you cant set the UA.

In such a (not uncommon) setup "resty" would show up as the UA which is not helpful at all. The user of resty already supplies a UA, there is no need to add another header with the resty UA.

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -79,8 +79,6 @@ func parseRequestHeader(c *Client, r *Request) error {
 
 	if IsStringEmpty(hdr.Get(hdrUserAgentKey)) {
 		hdr.Set(hdrUserAgentKey, fmt.Sprintf(hdrUserAgentValue, Version))
-	} else {
-		hdr.Set("X-"+hdrUserAgentKey, fmt.Sprintf(hdrUserAgentValue, Version))
 	}
 
 	if IsStringEmpty(hdr.Get(hdrAcceptKey)) && !IsStringEmpty(hdr.Get(hdrContentTypeKey)) {


middleware.go
```diff
@@ -79,8 +79,6 @@ func parseRequestHeader(c *Client, r *Request) error {
 
 	if IsStringEmpty(hdr.Get(hdrUserAgentKey)) {
 		hdr.Set(hdrUserAgentKey, fmt.Sprintf(hdrUserAgentValue, Version))
-	} else {
-		hdr.Set("X-"+hdrUserAgentKey, fmt.Sprintf(hdrUserAgentValue, Version))
```
jeevatkm: @Robbilie Thank you for the PR. I have read your explanation and understood. 

Instead removing, please change to following:
```go
else {
    hdr.Add(hdrUserAgentKey, fmt.Sprintf(hdrUserAgentValue, Version))
}
```

Robbilie: what does that do? Append to the UA? If so I would also advice against this since it leaks information about the client/version (just like you shouldnt expose your webservers or os version on publicly visible status pages).

jeevatkm: Yes, it appends to it. For server I agree, we shouldn't expose it. But server should know about the client details.

Robbilie: I dont want to have that part included in my UA and i would leave that option to the user to include it. If you want to give the user the option, feel free to add a config flag to enable the appending to the UA, I wont and I cant since i am not a go programmer.

jeevatkm: Fair enough. I will merge this PR. Thank you for your time. Yes, if user want they can add it.

RenathoAzevedo:
@jeevatkm Could you set an option to disable the default UA header? I would like not to set the UA and I don't wanna have the:

> User-Agent: go-resty v1.0 - https://github.com/go-resty/resty
is it possible or already implemented?

jeevatkm:
@RenathoAzevedo you can set your own user agent header. If you provide one resty will not set default UA.

RenathoAzevedo:
@jeevatkm yes, I'm aware of it, but I would like not to have one set and neither default.

jeevatkm:
@RenathoAzevedo Okay, currently no settings to exclude default UA header however please make use of `SetPreRequestHook` to remove UA header. 

Will add an optional boolean flag to disable default UA. Can you create an issue for this this option?

Robbilie:
@RenathoAzevedo not setting a UA is bad practice, just saying…

PR #93, opened 2017-09-06
add NewRequest() method as alias for R()

kovetskiy:
Hi, I would like to use your package but I prefer to use more obvious name of
function.

Name of function must describe what function does.


kovetskiy:
@jeevatkm I've force pushed commit, all checks have passed.

jeevatkm:
@kovetskiy Thank you.

PR #99, opened 2017-10-17
check response size for debug logs

sudo-suhas:
In my case, for some endpoints, the response was too large(ex: length - 1616599). This made it very difficult to use debug mode, which btw, is really awesome. This PR checks the size of the response and if it exceeds 1000000, `***** RESPONSE TOO LARGE *****` is logged instead.


response.go
```diff
@@ -120,19 +120,22 @@ func (r *Response) RawBody() io.ReadCloser {
 }
 
 func (r *Response) fmtBodyString() string {
-	bodyStr := "***** NO CONTENT *****"
 	if r.body != nil {
+		if len(r.body) > 1000000 {
```
jeevatkm: What is this `1000000` value?  are you picking random max value to limit it? and why?

response.go
```diff
@@ -120,19 +120,22 @@ func (r *Response) RawBody() io.ReadCloser {
...
+			return "***** RESPONSE TOO LARGE *****"
+		}
 		ct := r.Header().Get(hdrContentTypeKey)
 		if IsJSONType(ct) {
 			out := acquireBuffer()
 			defer releaseBuffer(out)
-			if err := json.Indent(out, r.body, "", "   "); err == nil {
-				bodyStr = string(out.Bytes())
+			err := json.Indent(out, r.body, "", "   ")
```
jeevatkm: This could be better and concise.

```go
if err := json.Indent(out, r.body, "", "   "); err == nil {
    return out.String()
}
```

jeevatkm:
Also test coverage is decreasing, spend some time on it. Thank you.

response.go
sudo-suhas: Yea.. It is arbitrary. What do you suggest? Should I add a `var` so that the user can override if required?

response.go
sudo-suhas: That results in a warning from golint - `if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (golint)`.

sudo-suhas:
Could you tell me where I would need to add tests? I did not see `response_test.go` nor are there any tests for `fmtBodyString`.

response.go
jeevatkm: Making this value configurable at [client level](https://github.com/go-resty/resty/blob/master/client.go#L68) would be best. Since some user wants to get complete body value.

* Add an un-exported field `debugBodySizeLimit int64` to `Client` and method in the client `SetDebugBodyLimit(sl int64)`
* Initializing `debugBodySizeLimit` default value as `math.MaxInt32` (this is good enough to begin with) at [client instance creation](https://github.com/go-resty/resty/blob/master/default.go#L27) 
* Add corresponding method for default resty client in [default.go](https://github.com/go-resty/resty/blob/master/default.go)

This way it will satisfy possibilities.

response.go
jeevatkm: Actually remove this line `return "unexpected error: " + err.Error()` with my above code snippet then golint warning goaway. Because if any errors in formatting debug log, return the original body value which is `r.String()`.

jeevatkm:
To improve test you need create a scenario for response body limit. With configurable value you can decrease the value to create a scenario.

You can add the test in `client_test.go` or `resty_test.go`. Just focus on your changes/implementation and compare with master branch code coverage number.

Use following commands to know coverage lines info:
```
go test -coverprofile=coverage.out -race
go tool cover -html=coverage.out
```

jeevatkm:
@sudo-suhas Thank for your contribution and improving tests.

PR #101, opened 2017-10-20
Add GetClient method

jfarrell:
Adds a GetClient method to return the httpClient currently set.

### DIFF
--- a/client.go
+++ b/client.go
@@ -716,6 +716,11 @@ func (c *Client) IsProxySet() bool {
 	return c.proxyURL != nil
 }
 
+// GetClient method returns the current http.Client used by the resty client.
+func (c *Client) GetClient() *http.Client {
+	return c.httpClient
+}
+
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -716,6 +716,11 @@ func (c *Client) IsProxySet() bool {
 	return c.proxyURL != nil
 }
 
+// GetClient method returns the current http.Client used by the resty client. 
+func (c *Client) GetClient() *http.Client {
```
jeevatkm: The same `GetClient` method required for default client too at this file [default.go](https://github.com/go-resty/resty/blob/master/default.go). Could you please add it?

```go
// GetClient method returns the current `http.Client` used by the default resty client.
func GetClient() *http.Client {
  return DefaultClient.httpClient
}
```

And adding test coverage for these two methods.

jfarrell: +1, can do

jeevatkm: Thank you @jfarrell

jfarrell:
@jeevatkm updated default.go and added test cases, let me know if you would like anything else for this. Thanks for reviewing

jeevatkm:
@jfarrell Thank you.

Issue #56, opened 2017-02-09
httpClient.Do is protected by a mutex?


wanghq:
Because of the locking, there can only be __one__ outgoing call, which is really bad to the performance. If I understand it correctly, 
* if req1 has proxyURL set but req2 doesn't, then req2 will be sent with req1's proxy. Is this expected?
* if req1 and req2 both have proxyURL set, then req1 will be sent with req1's proxy and req2 will be sent with req2's proxy. This works. But the overhead of this is to not have more than 1 outgoing calls.
* if `Unlock()` is moved up before httpClient.Do. For the above case, there is no guarantee on which proxy will be used, which is not good. 

I hope this per request proxy is an optional feature. Providing it is fine but without sacrificing the normal use case.

Code ref:
https://github.com/go-resty/resty/blob/master/client.go#L634-L647

```
	c.mutex.Lock()

	if req.proxyURL != nil {
		c.transport.Proxy = http.ProxyURL(req.proxyURL)
	} else if c.proxyURL != nil {
		c.transport.Proxy = http.ProxyURL(c.proxyURL)
	}

	req.Time = time.Now()
	c.httpClient.Transport = c.transport

	resp, err := c.httpClient.Do(req.RawRequest)

        c.mutex.Unlock()
```

jeevatkm:
I understand what you're describing. I have introduced locking in #29. but I did not [satisfied myself](https://github.com/go-resty/resty/issues/29#issuecomment-222594301).

It seems I have to make a call before v1.0 release on this locking.

jeevatkm:
@wanghq mutex removed, request level proxy setting is removed.  Locking is not needed anymore for resty.

Refer #29 too.

jeevatkm:
Updates are present in #29. Closing this one.

PR #107, opened 2017-12-04
remove Request level proxy

wppurking:
according to #56  the Request Level Proxy is removed.


jeevatkm:
@wppurking You're correct, thanks for updating documentation.

PR #111, opened 2017-12-21
Added Content-Type support for SetFileReader

gh0st42:
Simple solution to set content type for multipart filereader parts. Default resty behaviour of "application/octet-stream" still present if no content-type is given.

### DIFF
--- a/client.go
+++ b/client.go
@@ -85,19 +85,18 @@ type Client struct {
 	JSONMarshal           func(v interface{}) ([]byte, error)
 	JSONUnmarshal         func(data []byte, v interface{}) error
 
-	httpClient         *http.Client
-	setContentLength   bool
-	isHTTPMode         bool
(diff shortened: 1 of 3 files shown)


jeevatkm:
@gh0st42 Thanks for the PR. While looking at your PR, I think we can do better for `SetFileReader` and `SetFile` together with backward compatibility.

Let me create a PR, then you can have a look and we can improve it.

gh0st42:
Sounds good, if you have a special testing branch this would also be fine and I'd be happy to test it. I guess for a general solution that can be applied to `SetFileReader` as well as `SetFile` a bit more work has to be done since currently they are implemented both a bit different in resty.

jeevatkm:
Closing this PR infavour of generic implementation 513d3b071ad42983368d7d5ff102ab691e99c371

PR #121, opened 2018-02-04
Allow the Request to override the Authorization header set by the Client

arun251:
We have code that sets the `Authorization` header on the `Client` object and then overrides the header on the `Request` object.  That previously worked with Resty v0.13 but no longer worked after upgrading to Resty v1.2. The root cause of this issue is in `parseRequestHeader` where client and request headers are concatenated.

This PR fixes the issue for `Authorization`, but I suspect there are other singleton headers that could be affected.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -472,3 +472,20 @@ func TestSetLogPrefix(t *testing.T) {
 	assertEqual(t, "CUSTOM ", c.logPrefix)
 	assertEqual(t, "CUSTOM ", c.Log.Prefix())
 }
+
+func TestRequestOverridesClientAuthorizationHeader(t *testing.T) {
+	ts := createAuthServer(t)
+	defer ts.Close()
+
+	c := dc()
+	c.SetTLSClientConfig(&tls.Config{InsecureSkipVerify: true}).
+		SetHeader("Authorization", "some token")
+		SetHostURL(ts.URL + "/")
+
+	resp, err := c.R().
+		SetHeader("Authorization", "Bearer 004DDB79-6801-4587-B976-F093E6AC44FF").
+		Get("/profile")
+
+	assertError(t, err)
+	assertEqual(t, http.StatusOK, resp.StatusCode())
+}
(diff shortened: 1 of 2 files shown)


jeevatkm:
@arun251 Thank you for the PR. 

I will verify this behaviour between v0.13 and latest. I good to ensure any other singleton header also is not affected.

jeevatkm:
@arun251 If there is an issue with singleton headers, that needs to be fixed across not just Authorization. Thanks for your findings. I will analyze and get back to you.

client_test.go
```diff
@@ -472,3 +472,20 @@ func TestSetLogPrefix(t *testing.T) {
 	assertEqual(t, "CUSTOM ", c.logPrefix)
 	assertEqual(t, "CUSTOM ", c.Log.Prefix())
 }
+
+func TestRequestOverridesClientAuthorizationHeader(t *testing.T) {
```
jeevatkm: Please move this test case to `request_test.go` file.

middleware.go
```diff
@@ -76,6 +76,11 @@ func parseRequestHeader(c *Client, r *Request) error {
 		hdr[k] = append(hdr[k], r.Header[k]...)
 	}
 
+	// Allow the Request to override the Authorization header set by the Client
```
jeevatkm: I have analyzed the v0.13 and latest version. Fix for across headers.

Remove this if condition and update your PR as follows.

Update [Line #75 to #77](https://github.com/go-resty/resty/blob/master/middleware.go#L75-L77) to following-  
```go
for k := range r.Header {
	hdr.Del(k)
	hdr[k] = append(hdr[k], r.Header[k]...)
}
```

jeevatkm:
@arun251 Awaiting for your PR changes or let me know I will do it.

arun251:
Please go ahead with the changes. Thanks!

jeevatkm:
@arun251 Thank you, I will take your test case and create new PR.

PR #122, opened 2018-02-04
Allow uncommon HTTP methods to support payloads

arun251:
We use `REPORT` and [other HTTP methods](https://www.iana.org/assignments/http-methods/http-methods.xhtml) as part of our REST API for their defined semantics. The `REPORT` method is defined to support a request body. I've refactored `isPayloadSupported` to explicitly look for the standard HTTP methods that are defined as not supporting request bodies so that uncommon HTTP methods may support request bodies, too.

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -1246,3 +1246,17 @@ func TestGetPathParams(t *testing.T) {
 		"userId": "sample@sample.com",
 	})
 }
+
+func TestReportMethodSupportsPayload(t *testing.T) {
+	ts := createGenServer(t)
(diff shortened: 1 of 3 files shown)


jeevatkm:
@arun251 Thank you for the PR.

Do you think, is it good to add HTTP verb method for `Report` like [other methods](https://github.com/go-resty/resty/blob/master/request.go#L388-L420)?

arun251:
It would be nice, but I'm equally happy using `Execute("REPORT", url)`.

PR #141, opened 2018-04-11
Add support for bazel build rules

paradoxengine:
Adding a basic workspace and bazel rules.
The rules were generated via gazelle (blaze run :gazelle) and tuned a bit (e.g. changing visibility and adding the data dependency).
This should allow bazel users to just depend on the project without a local BUILD file.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,47 @@
+package(default_visibility = ["//visibility:private"])
+
+load("@bazel_gazelle//:def.bzl", "gazelle")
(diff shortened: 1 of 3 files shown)


paradoxengine:
Sure, I've added some simple text, let me know if that's enough.
Bazel is a build system opensourced by Google: it manages dependencies differently than go and requires to have the WORKSPACE/BUILD files for folks to depend on it. The alternative is to essentially have the BUILD file on the caller side, which is not so great.

PR #158, opened 2018-06-07
Typo in example, client1 should be client2

intel352:


jeevatkm:
Thanks for improving doc.

PR #174, opened 2018-08-10
Make IsJSONType and IsXMLType more robust

moorereason:
The existing IsJSONType and IsXMLType regular expressions are here
generalized to work with any properly formatted media type.


moorereason:
Attempting to generalize the changes in #172.

Issue #181, opened 2018-08-29
Fix resty for go mod tidy and go mod why


jeevatkm:
Issue details is here https://github.com/go-resty/resty/issues/178#issuecomment-417116684

Fix and make a release of `v1.9.1`

jeevatkm:
@vvarp @zuf I have released `v1.9.1`. Could you please try out and let me know?

Issue #178, opened 2018-08-23
Go Module


cimnine:
The documentation suggests that one should use resty as follows:

```
import "gopkg.in/resty.v1"
```

(https://github.com/go-resty/resty/blob/ec17de1c59a99e0743dac77aaab6435f364fdc42/README.md#usage)

But this brakes with Go modules, because `go.mod` declares the module as follows:

https://github.com/go-resty/resty/blob/ec17de1c59a99e0743dac77aaab6435f364fdc42/go.mod#L1

This causes troubles when I try to compile code with go11.1rc2 and go modules enabled.

All is fine when I import resty like this:

```go
import "github.com/go-resty/resty"
```

Now there are two options:

a) Adjust the documentation and suggest to `import "github.com/go-resty/resty"`
b) Adjust the `go.mod` to be `module "gopkg.in/resty.v1`

Option (a) is good for all new project that start with go modules right away, but it's backward incompatible.
Option (b) would be good for projects that used plain `go get` or `dep` until now and upgrade to `go.mod`, because they don't have to change all the import statements. There are no further drawbacks, except that this option needs a new release with an updated `go.mod`.

jeevatkm:
@cimnine I would like to clarify if understood correctly.

You mean to say using `gopkg.in/resty.v1` with Go modules breaks the build, correct?

----

By nature of resty package it will work seamlessly with `github.com/go-resty/resty` and `gopkg.in/resty.v1` in plain `go get` (on go1.11 it has to have `GO111MODULE=off`). It keeps backward compatibility with plain go get. Could you please try and let me know?

But if we use `gopkg.in/resty.v1` with Go modules it will break, as you have mentioned `go.mod` defines it as `github.com/go-resty/resty`. This is expected behavior.

FYI, resty package currently does not enforce package import like `package resty // import "github.com/go-resty/resty"`. I'm planning to do that for `v2.0.0` release in favor of Go modules.

I understand, documentation on `Usage` section needs update for Go modules instructions. I will do (or you can also send PR).  Thank you for bringing it up.

cimnine:
> You mean to say using gopkg.in/resty.v1 with Go modules breaks the build, correct?

Yes.

> github.com/go-resty/resty

This will work with plain `go get` as well.

> I understand, documentation on Usage section needs update for Go modules instructions.

Yes.

[...]

PR #190, opened 2018-09-28
Update go.mod

leonsim:
It is mandatory to omit `v0` or `v1` according to `Why are v0, v1 omitted in the import paths? Why must the others appear? Why must v0, v1 never appear?` in https://github.com/golang/go/issues/24301

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module gopkg.in/resty.v1
+module gopkg.in/resty
 
 require golang.org/x/net v0.0.0-20180911220305-26e67e76b6c3


leonsim:
Yes, it does work with without `v1` definition in the `go.mod` with `gopkg.in`.
To prove this, I push my changes to the master branch of my folked repo, and create a new tag `v1.9.2`. Now you can see this new tag in  https://gopkg.in/leonsim/resty.v1

Actually, I created this PR because I got an error when I run `go mod tidy` command in one of my projects. It complains not able to parse the `go.mod` file in `resty` repo.

jeevatkm:
@leonsim Thanks for the details. I would like to keep go mod definition as is till v2 release without impacting existing users (without changing import path). More info #178 & #181 around go mod.

---

> Actually, I created this PR because I got an error when I run go mod tidy command in one of my projects. It complains not able to parse the go.mod file in resty repo.

This is strange, I have actually verified with `v1.9.1` no issues with any commands from `go mod` (#178 & #181). Could you please clear your local copies and try v1.9.1 and let me know.

leonsim:
After I update `go.mod` to `require gopkg.in/resty.v1 v1.9.1` and then `go mod tidy` works without any error.

Issue #206, opened 2019-01-03
Proposal: Add Context() method into Request and consider logging prerequesthook updates


clippit:
Hi,

Firstly thank you for creating such a great tool! 😄 

I find it's inconvenient for passing user custom parameters when using user-defined middleware. For example, in our development process, the server may receive an `X-Request-ID` header and it should pass this header to any services it calls. That's a simple way for tracing RPC calling actually. Besides of this, we also have some requirements for generating signature per request.

For current version of resty, the only way to approach this is using `SetPreRequestHook` and carrying user values with `req.RawRequest.Context()`, e.g:

```
client.SetPreRequestHook(func(c *resty.Client, req *resty.Request) error {
	requestId, ok := req.RawRequest.Context().Value(requestIdKey).(string)
	if ok && requestId != "" {
		req.SetHeader("x-request-id", requestId)
	}

	authKey, ok := req.RawRequest.Context().Value(authKey).(string)
	if !ok {
		return errors.New("no auth key in request context")
	}

	// ... grab request info to generate signature...
	req.SetHeader("x-signature", "...")

	return nil
})
```

When using this client, we need a nested context to store these values:

```
ctx := context.WithValue(context.Background(), requestIdKey, "from_current_server_request_header")
ctx = context.WithValue(ctx, authKey, "may_returned_by_another_service")
client.R().SetContext(ctx).Get("/")
```

We can only use `SetPreRequestHook` rather than `OnBeforeRequest` because `RawRequest` is created after all `OnBeforeRequest` middlewares and the context of RawRequest if the only place that can carry these values. But there are some shortcoming:

1. The headers added in `SetPreRequestHook` cannot be printed when debug is on since `requestLogger` happens before this hook
2. It's not so good to create so many context

So here is my modest proposal. A `GetContext()` method can be added into `Request` so we can put these codes in `OnBeforeRequest` and make request logger prints completed headers. Furthermore, the `Request` may contain a `map[string]interface{}` registry to store any kind of user values to get more abilities.

jeevatkm:
@clippit Thanks for bringing up your concerns and your appreciations. I understand your suggestions.

I will bring following updates:

- [ ] Expose context via `Request.Context()`
- [ ] Move request logger after `PreRequestHook` call

jeevatkm:
@clippit I have implemented in branch master. Could you try and let me know?

clippit:
It's good to me:) and I have added some checks, please take a look at #207

[...]

Issue #167, opened 2018-07-05
Resty v2.0.0 Action Item list


jeevatkm:
This thread is to keep track of action item comes out from #166 then respective issue would be created once its finalized for `v2.0.0`

Action Items:

- Improve documentation and examples
- Sane default timeouts for client and transport https://github.com/go-resty/resty/issues/166#issuecomment-402763015
- Add method into struct `Request` to accept type `url.Values` for Query strings  https://github.com/go-resty/resty/issues/166#issuecomment-406465002
- Pass error returned by `resty.Client.execute()` to retry condition callback https://github.com/go-resty/resty/issues/166#issuecomment-425523656
- Refactored resty to move away from `gopkg.in` and completely relay on `go mod` for packaging versioning
- Add on-demand tracer refer to https://github.com/go-resty/resty/issues/205#issuecomment-450756975 https://gist.github.com/moorereason/f77d69a2c823fb222eba61c151e63a6b
- Change PreRequestHook signature to `func(cl *Client, r *http.Request) error` - https://github.com/go-resty/resty/pull/207#issuecomment-451732759
- Log request and response debug log together for easy debugging

jeevatkm:
Resty v2 development issues gonna be created then this one will be closed 😄

jeevatkm:
Respective issues are created closing this one.

Issue #62, opened 2017-03-24
Add an option for user-defined middleware after the built-in middlware, to modify the RawRequest object


ccp-ccollins:
I wanted to use resty for requests to an AWS endpoint that requires requests to be signed with AWS credentials.  A library already exists for this (https://github.com/smartystreets/go-aws-auth), however using that one requires access to the http.Request object directly.  Currently, user defined middleware is run before the built-in middleware, which makes sense when modifying the resty.Request object.  However for resty to be compatible with go-aws-auth, I need middleware that is processed *after* the built-in middleware, so that the RawRequest object is available.

jeevatkm:
Okay I will add. Thanks.

jeevatkm:
@ccp-ccollins I thought through it, adding one more before request middleware will lead to confusion. Instead I'm planing to add `SetPreRequestHook` with same signature as middleware but only one can be added.

This will be available in `v0.12` release.

jeevatkm:
@ccp-ccollins Sorry for the delay. Usually I make it quick. This time I got occupied with my [aah framework](https://aahframework.org) first public release.

`SetPreRequestHook` is ready and will be release with `v0.12`.

*_PS: If you could spare sometime, please have a look on aah framework and share your inputs. I would help me to improve the framework._*

PR #207, opened 2019-01-04
point to the same context after http.Request created

clippit:
This pull request adds checks related to context methods. When http.Request is created, context between resty.Request and http.Request stays the same. This can avoid mess when working with preReqHook.

See #206

### DIFF
--- a/request17.go
+++ b/request17.go
@@ -55,6 +55,9 @@ type Request struct {
 // Context method returns the Context if its already set in request
 // otherwise it creates new one using `context.Background()`.
 func (r *Request) Context() context.Context {
+	if r.RawRequest != nil {
+		return r.RawRequest.Context()
+	}
 	if r.ctx == nil {
 		return context.Background()
 	}
@@ -67,6 +70,9 @@ func (r *Request) Context() context.Context {
 // documentation.
 func (r *Request) SetContext(ctx context.Context) *Request {
 	r.ctx = ctx
+	if r.RawRequest != nil {
+		r.addContextIfAvailable()
+	}
 	return r
 }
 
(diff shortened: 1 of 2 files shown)


request17.go
```diff
@@ -67,6 +70,9 @@ func (r *Request) Context() context.Context {
 // documentation.
 func (r *Request) SetContext(ctx context.Context) *Request {
 	r.ctx = ctx
+	if r.RawRequest != nil {
```
jeevatkm: @clippit I have intentionally did not make these changes. Because by resty design PreRequestHook is only meant to modify `RawRequest`  not resty Request` (refer to #62 ). Any changes made in resty Request instance will not affect the request.

request17.go
```diff
@@ -55,6 +55,9 @@ type Request struct {
 // Context method returns the Context if its already set in request
 // otherwise it creates new one using `context.Background()`.
 func (r *Request) Context() context.Context {
+	if r.RawRequest != nil {
```
jeevatkm: @clippit I have intentionally did not make these changes. Because by resty design PreRequestHook is only meant to modify `RawRequest`  not resty Request` (refer to #62). Any changes made in resty Request instance will not affect the request.

clippit:
I understand.
Only expose `RawRequest` in PreRequestHook may be a better option since whatever you want to get from `Request` can also be fetched from `RawRequest`, which also makes resty more opinionated about this hook. But this would be an API break change, perhaps you can consider it in next major version.

jeevatkm:
@clippit Thank you. Yes, I have realized when I was reviewing your PR. I will surely change hook signature in upcoming v2.  Just added as action item in #167

PR #211, opened 2019-01-09
 Add batch SetMultipartFields in order to set multiple fields

e11ni:
Dear Jeevanandam,

Thank you for your great library. I would like to propose you this small improvement.
The idea is to add SetMultipartFields method to set multiple fields. Sometimes it is convenient to prepare list with some hardcoded fields at top of request processing function in declarative way and then pass it to request. Resty API already has similar functions: SetHeaders and SetFiles.
Example of usage SetMultipartFields method (extracted from comment):
```
fields := []*MultipartField{
		&MultipartField{
			Param:       "uploadManifest1",
			FileName:    "upload-file-1.json",
			ContentType: "application/json",
			Reader:      strings.NewReader(`{"input": {"name": "Uploaded document 1", "_filename" : ["file1.txt"]}}`),
		},
		&MultipartField{
			Param:       "uploadManifest2",
			FileName:    "upload-file-2.json",
			ContentType: "application/json",
			Reader:      strings.NewReader(`{"input": {"name": "Uploaded document 2", "_filename" : ["file2.txt"]}}`),
		},
	}
resty.R().SetMultipartFields(fields)
```

### DIFF
--- a/request.go
+++ b/request.go
@@ -299,14 +299,14 @@ func (r *Request) SetMultipartField(param, fileName, contentType string, reader 
 
 // SetMultipartFields method is to set multiple data fields using io.Reader for multipart upload.
 // Example: set multiple json strings.
-//  fields := []*MultipartField{
-//		&MultipartField{
+//  fields := []*resty.MultipartField{
+//		&resty.MultipartField{
 //			Param:       "uploadManifest1",
 //			FileName:    "upload-file-1.json",
 //			ContentType: "application/json",
 //			Reader:      strings.NewReader(`{"input": {"name": "Uploaded document 1", "_filename" : ["file1.txt"]}}`),
 //		},
-//		&MultipartField{
+//		&resty.MultipartField{
 //			Param:       "uploadManifest2",
 //			FileName:    "upload-file-2.json",
 //			ContentType: "application/json",
--- a/request.go
+++ b/request.go
@@ -298,6 +298,22 @@ func (r *Request) SetMultipartField(param, fileName, contentType string, reader 
 }
 
 // SetMultipartFields method is to set multiple data fields using io.Reader for multipart upload.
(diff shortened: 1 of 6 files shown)


request.go
```diff
@@ -297,6 +297,31 @@ func (r *Request) SetMultipartField(param, fileName, contentType string, reader
...
+//		},
+//		&resty.MultipartField{
+//			Param:       "uploadManifest2",
+//			FileName:    "upload-file-2.json",
+//			ContentType: "application/json",
+//			Reader:      strings.NewReader(`{"input": {"name": "Uploaded document 2", "_filename" : ["file2.txt"]}}`),
+//		},
+//	}
+//  resty.R().SetMultipartFields(fields)
+func (r *Request) SetMultipartFields(fields []*MultipartField) *Request {
```
jeevatkm: @e11ni Could you change the func to `variadic` func `SetMultipartFields(fields ...*MultipartField)`? I think it would convenient to use.

e11ni:
I have changed signature to variadic and removed empty new lines.
I'm not sure about usage example comment before SetMultipartFields, probably it is better to show both usage examples: pass slice as argument and pass multiple fields one by one, but it will take many lines. @jeevatkm, need your advice regarding this comment.
Also I have noticed that some other functions have empty new lines. Could I remove them?

e11ni:
Comment looks good, will use it. About functions with empty new lines:
SetFile
SetContentLength
Not sure but probably it also applicable for following (as they use just 3-4 statements in the row without any loops and if-statements):
SetFileReader
SetMultipartField

jeevatkm:
@e11ni I just had a look after your comment. Yes they have it. Is it possible for you to remove those lines?

e11ni:
@jeevatkm I have removed empty new lines in those four functions.

Issue #225, opened 2019-02-12
Potential race in resty with combination of retries and the buffers pool


amits1995:
We have a golang project that uses resty as the http client.  
We started using the golang race detector in our staging environment and encountered the following race errors:
```
==================
WARNING: DATA RACE
Write at 0x00c00025e610 by goroutine 149:
  bytes.(*Buffer).Read()
      /usr/local/go/src/bytes/buffer.go:299 +0x4a
  io/ioutil.(*nopCloser).Read()
      <autogenerated>:1 +0x87
  net/http.(*http2clientStream).writeRequestBody()
      /usr/local/go/src/net/http/h2_bundle.go:7825 +0x635
  net/http.(*http2Transport).getBodyWriterState.func1()
      /usr/local/go/src/net/http/h2_bundle.go:9040 +0x13d

Previous write at 0x00c00025e610 by goroutine 109:
  bytes.(*Buffer).Read()
      /usr/local/go/src/bytes/buffer.go:299 +0x4a
  io/ioutil.(*nopCloser).Read()
      <autogenerated>:1 +0x87
  net/http.(*http2clientStream).writeRequestBody()
      /usr/local/go/src/net/http/h2_bundle.go:7825 +0x635
  net/http.(*http2Transport).getBodyWriterState.func1()
      /usr/local/go/src/net/http/h2_bundle.go:9040 +0x13d

Goroutine 149 (running) created at:
  net/http.http2bodyWriterState.scheduleBodyWrite()
      /usr/local/go/src/net/http/h2_bundle.go:9087 +0xb8
  net/http.(*http2ClientConn).roundTrip()
      /usr/local/go/src/net/http/h2_bundle.go:7626 +0x81b
  net/http.(*http2Transport).RoundTripOpt()
      /usr/local/go/src/net/http/h2_bundle.go:7037 +0x22d
  net/http.(*http2Transport).RoundTrip()
      /usr/local/go/src/net/http/h2_bundle.go:6999 +0x4b
  net/http.(*Transport).roundTrip()
      /usr/local/go/src/net/http/transport.go:463 +0xa21
  net/http.(*Transport).RoundTrip()
      /usr/local/go/src/net/http/roundtrip.go:17 +0x42
  net/http.send()
      /usr/local/go/src/net/http/client.go:250 +0x304
  net/http.(*Client).send()
      /usr/local/go/src/net/http/client.go:174 +0x1ca
  net/http.(*Client).do()
[...]

PR #228, opened 2019-02-20
Add request body race test

moorereason:
This PR adds a test case for the race condition reported in #225 and adds the race detector to Travis CI.

No fix for the race is included yet.  I wanted to get feedback first.

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -16,6 +16,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


jeevatkm:
@moorereason I will try this PR on my local and get back to you.

request_test.go
```diff
@@ -28,6 +29,41 @@ type AuthError struct {
 	ID, Message string
 }
 
+func TestRequestRace(t *testing.T) {
+	// Issue #225
+
+	ts := createGetServer(t)
+	defer ts.Close()
+
+	req := R()
```
jeevatkm: @moorereason I have looked at this PR and tried out locally. This is incorrect usage because this test case sharing `Request` object across goroutine. Resty does not work that way. Each request processed in the dedicated request object. 

Usage is -
```go
req := R()
resp, err := req.Get(ts.URL + "/")
```

And #225 issue involves the Request body and user reported data race in request body writing.

moorereason: I agree with you.  I mistakenly thought this was related to #225 when I submitted it (in haste).  I only found this PR while trying to create data races with resty.  I don't know that anyone would actually do this in production if they understand Go concurrency.

My only mild suggestion would be to clarify the docs to say that the "goroutine concurrent safe" guarantee applies on to the resty Client.

Closing this PR.

PR #234, opened 2019-03-13
fix multipart/form-data without filename and content type

larryhu:
Sometime just want send fields without file when content-type is multipart/form-data

### DIFF
--- a/util.go
+++ b/util.go
@@ -139,9 +139,17 @@ func escapeQuotes(s string) string {
 
 func createMultipartHeader(param, fileName, contentType string) textproto.MIMEHeader {
 	hdr := make(textproto.MIMEHeader)
-	hdr.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"; filename="%s"`,
-		escapeQuotes(param), escapeQuotes(fileName)))
-	hdr.Set("Content-Type", contentType)
+	param = escapeQuotes(param)
+	hdr.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"`, param))
+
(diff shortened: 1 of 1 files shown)


util.go
```diff
@@ -139,9 +139,17 @@ func escapeQuotes(s string) string {
 
 func createMultipartHeader(param, fileName, contentType string) textproto.MIMEHeader {
 	hdr := make(textproto.MIMEHeader)
-	hdr.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"; filename="%s"`,
-		escapeQuotes(param), escapeQuotes(fileName)))
-	hdr.Set("Content-Type", contentType)
+	param = escapeQuotes(param)
```
jeevatkm: @larryhu Could you please target your PR against `v2` branch? I'm planning to freeze the v1 series with `v1.12.0`. 

Also, could please update the code into following?

```go
func createMultipartHeader(param, fileName, contentType string) textproto.MIMEHeader {
	hdr := make(textproto.MIMEHeader)
	var contentDispositionValue string
	if IsStringEmpty(fileName) {
		contentDispositionValue = fmt.Sprintf(`form-data; name="%s"`, param)
	} else {
		contentDispositionValue = fmt.Sprintf(`form-data; name="%s"; filename="%s"`,
			param, escapeQuotes(fileName))
	}
	hdr.Set("Content-Disposition", contentDispositionValue)

	if !IsStringEmpty(contentType) {
		hdr.Set("Content-Type", contentType)
	}
	return hdr
}
```

larryhu:
ok, thanks. and...  close this PR ?

PR #237, opened 2019-03-19
change Backoff() algorithm

neganovalexey:
1) Add capabilities to handle Retry-After headers and similar info from
server

Motivation: some servers provide Retry-After header or similar info along with 429
or 503 status code, and it is often important to honor such information
on retries, i.e. simple expotential backoff is not optimal.
https://docs.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online

2) Add option NOT to retry even if operation returns an error (but retry
by default, if no retry conditions are set)

Motivation: error are already passed to condition callback in resty, but
Backoff() still retries the request if error is not nil. It implies excessive,
stillborn retries for non-retryble errors from underlying http client
(i.e. with RoundTripper from oauth2).

3) Remove error return value from condition callback

Motivation: this error is neither passed to caller, nor logged in any
way. It is cleaner to have "needRetry == true" than "needRetry == true
|| conditionErr != nil".

4) Does not use floating-point arithmetics for expotential backoff

Motivation: simplification & performance

### DIFF
--- a/client.go
+++ b/client.go
@@ -89,6 +89,7 @@ type Client struct {
 	RetryWaitTime         time.Duration
 	RetryMaxWaitTime      time.Duration
 	RetryConditions       []RetryConditionFunc
+	RetryAfter            RetryAfterFunc
 	JSONMarshal           func(v interface{}) ([]byte, error)
 	JSONUnmarshal         func(data []byte, v interface{}) error
 
@@ -513,6 +514,13 @@ func (c *Client) SetRetryWaitTime(waitTime time.Duration) *Client {
 func (c *Client) SetRetryMaxWaitTime(maxWaitTime time.Duration) *Client {
 	c.RetryMaxWaitTime = maxWaitTime
 	return c
+}
+
+// SetRetryAfter sets callback to calculate wait time between retries.
+// Default (nil) implies exponential backoff with jitter
+func (c *Client) SetRetryAfter(callback RetryAfterFunc) *Client {
+       c.RetryAfter = callback
+       return c
 }
 
 // AddRetryCondition method adds a retry condition function to array of functions
--- a/request.go
+++ b/request.go
@@ -566,7 +566,7 @@ func (r *Request) Execute(method, url string) (*Response, error) {
 
 	var resp *Response
 	attempt := 0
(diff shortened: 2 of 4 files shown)


neganovalexey:
I will glad to receive any comments!
Maybe it will be nice to split the commit into several ones, but I think it is acceptable to show the draft in one piece.

jeevatkm:
@neganovalexey Also, could you have a look at the test coverage? which edge case is missing from the unit test.

pborzenkov:
> Add option NOT to retry even if the operation returns an error (but retry by default, if no retry conditions are set)
> [Jeeva] This is a kind of design choice, let's introduce this behavior and wait for user feedback to re-evaluate in the future on behavior.

Unfortunately, if request is always retried in the presence of non-nil error, passing this err to retry condition callback doesn't make much sense, as a user can't disable retry for particular kinds of errors (e.g. invalid token, etc.).

neganovalexey:
> @neganovalexey Also, could you have a look at the test coverage? which edge case is missing from the unit test.

done

neganovalexey:
@jeevatkm , floating-point arithmetics is less precise than integer arithmetics. It doesn't matter for this code, the corresponding changes were done mainly for aesthetic reasons (it is painful to see avoidable floating-point operations for a person with some C background).
I understand that it looks like some unnecessary refactoring and I'm going to return the old piece of code with math.Exp2() call.

jeevatkm:
> Unfortunately, if request is always retried in the presence of non-nil error, passing this err to retry condition callback doesn't make much sense, as a user can't disable retry for particular kinds of errors (e.g. invalid token, etc.).

@pborzenkov Its not always retried when error is non-nil. My understanding is - it happens only when retry count >0 and retry condition is not set.

@neganovalexey could you please validate my understanding and respond to @pborzenkov?

pborzenkov:
@jeevatkm I was referring to the original way the code was written. Merged patch definitely allows to stop retry for particular kinds of errors, which is good! Thanks!

jeevatkm:
@pborzenkov Thank you 😄

PR #259, opened 2019-08-05
exposes Close() method and set response body to nil after read

divinerapier:


client.go
```diff
@@ -805,7 +805,11 @@ func (c *Client) execute(req *Request) (*Response, error) {
 	}
 
 	if !req.isSaveResponse {
-		defer closeq(resp.Body)
+		defer func() {
+			closeq(resp.Body)
+			// response body should be used no longer
+			resp.Body = nil
```
jeevatkm: @divinerapier Is this `nil` assignment necessary after closing the body, can you please explain?

response.go
```diff
@@ -134,6 +134,14 @@ func (r *Response) IsError() bool {
 	return r.StatusCode() > 399
 }
 
+// Close method closes the inner HTTP response body.
+func (r *Response) Close() error {
```
jeevatkm: @divinerapier `r.RawBody()` returns value only if Resty user uses https://godoc.org/github.com/go-resty/resty#Request.SetDoNotParseResponse otherwise it will be always `nil` since resty fully reads the response body and closes it. 

Can you please explain or I might be failing to understand something?

jeevatkm:
@divinerapier Follow up. May I know the issues you have faced and the reason for the changes in this PR?

divinerapier:
I'm so sorry, this pr is too long to remember it.

PR #277, opened 2019-10-09
Added capacity to slice creation, when capacity is known

vivekv96:
Simple optimisation to reduce memory allocations and copying(while appending).


vivekv96:
@jeevatkm Thank you.

PR #284, opened 2019-11-17
fix trace context overriding request context

xixinjie:


xixinjie:
You're very welcome, thank you for sharing this awesome lib!

Issue #289, opened 2019-11-22
Retries carry error to a success


phillc:
Hello,

I have experienced this in my code base as have translated it to a test in the code base.

In this test case, I have an error (should be anything >399) on a json request with SetError for the first request, and then a successful second request.

I also have a retry condition that is looking for the first error

The result of the execution has the body of the second request, but the error of the first request.

I initially thought this was due to the err variable https://github.com/phillc/resty/blob/master/request.go#L611-L618 being shadowed, but fixing that did not fix the problem. I am currently in the process of debugging `Backoff` in hopes of finding it, but am currently not successful.

Any thoughts into if my usage is just wrong, or where the error likely is would be helpful.
Thanks.
Pull request with test coming.

phillc:
I think this solves the problem, but I'm not 100% sure this is the right solution. https://github.com/go-resty/resty/pull/290/commits/6572463ad1bf3dd920f392c7f228b33951f01c1c

jeevatkm:
@phillc It's a correct spot of resetting an error, the same request object is used for retry. Thanks, PR #209 is merged!

PR #290, opened 2019-11-22
Failing first request and successfully retrying carries error from first

phillc:
Proves https://github.com/go-resty/resty/issues/289

### DIFF
--- a/retry_test.go
+++ b/retry_test.go
@@ -582,13 +582,12 @@ func TestClientRetryErrorRecover(t *testing.T) {
 
 	assertError(t, err)
 
-	responseErr := resp.Error().(*AuthError)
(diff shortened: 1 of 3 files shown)


retry_test.go
```diff
@@ -559,6 +559,38 @@ func TestClientRetryPost(t *testing.T) {
...
+
+	assertError(t, err)
+
+	responseErr := resp.Error().(*AuthError)
+	authSuccess := resp.Result().(*AuthSuccess)
+
+	assertEqual(t, http.StatusOK, resp.StatusCode())
+	assertEqual(t, "hello", authSuccess.Message)
+
+	assertNil(t, responseErr)
```
phillc: This last assertion is the one that fials

PR #313, opened 2020-02-20
added SetRootCertificateFromString method to read cert from string

dzaytsev91:
Now it's impossible to set root certificate from string value, only from file path


jeevatkm:
@dzaytsev91 Also, can you please increase test case coverage for the newly added method?

dzaytsev91:
> @dzaytsev91 Also, can you please increase test case coverage for the newly added method?

Sure, please review

PR #325, opened 2020-03-09
force option to attempt http2

golgoth31:
Resty does not use http2 golang conn. Forcing attempt option in transport allow resty to use http2

### DIFF
--- a/client.go
+++ b/client.go
@@ -24,8 +24,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"golang.org/x/net/http2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


golgoth31:
Hi, Thank you for your answer.
I'm ok to say that, by default, GO HTTP client is http/2 aware but, in the resty client definition, you create a totally new transport layer based on the Transport struct. This layer is not http/2 aware by default, leading your client not using http/2. It's not really clear but it's explained here: https://golang.org/src/net/http/transport.go

> // ForceAttemptHTTP2 controls whether HTTP/2 is enabled when a non-zero
> // Dial, DialTLS, or DialContext func or TLSClientConfig is provided.
> // By default, use of any those fields conservatively disables HTTP/2.
> // To use a custom dialer or TLS config and still attempt HTTP/2
> // upgrades, set this to true.

I wrote a simple main file:
```
package main

import (
	"fmt"
	"net/http"
)

func main() {
	client := resty.New()
	rresp, _ := client.R().Get("https://httpbin.org/ip")
	fmt.Printf("resty response protocol: %s\n", rresp.RawResponse.Proto)

	httpclient := http.DefaultClient
	hresp, _ := httpclient.Get("https://httpbin.org/ip")
	fmt.Printf("net/http response protocol: %s\n", hresp.Proto)
}
```

The result is as follow:
- without my patch:
❯ go run test.go
resty response protocol: HTTP/1.1
net/http response protocol: HTTP/2.0
- with my patch:
❯ go run test.go
resty response protocol: HTTP/2.0
net/http response protocol: HTTP/2.0

As shown, the current resty client is not http/2.

For the CI problem, between 1.12 and 1.13, the http2 transport enabling has been changed. I will investigate on that.

client.go
```diff
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"golang.org/x/net/http2"
```
jeevatkm, line 28: @golgoth31 Thanks for submitting PR. And providing insight into the issue. 
`golang.org/x/net/http2` package is part of the language since go1.6. So externally referring is not required.

I will investigate http2 and address it. Thanks again.

jeevatkm:
@golgoth31 Just now I have analyzed it using your code snippet. 
* It seems HTTP 2.0 broken `go1.13` onwards 
* It was working until `go1.12` 
* So, it's clearly evident it was working up to go1.12 so I did not go further.

**Results are -**

**Broken on go.14**
Screen Shot 2020-05-10 at 10 10 45 PM

**Broken on go.13**
Screen Shot 2020-05-10 at 10 11 00 PM

**Success on go.12**
Screen Shot 2020-05-10 at 10 11 29 PM

**Success on go.11**
Screen Shot 2020-05-10 at 10 11 42 PM

jeevatkm:
@golgoth31 Closing this PR in-favor of #339

PR #335, opened 2020-05-04
Fixed Debugf call in Request/Response logguer

phsym:
It was calling `Debugf(debugLog)` which tries to handle `%` parameters, resulting in strange logs. Must be called this way instead `Debugf("%s", debugLog)`

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -325,7 +325,7 @@ func responseLogger(c *Client, res *Response) error {
 		}
 		debugLog += "==============================================================================\n"
 
-		c.log.Debugf(debugLog)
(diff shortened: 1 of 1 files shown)


middleware.go
```diff
@@ -325,7 +325,7 @@ func responseLogger(c *Client, res *Response) error {
 		}
 		debugLog += "==============================================================================\n"
 
-		c.log.Debugf(debugLog)
+		c.log.Debugf("%s", debugLog)
```
jeevatkm, line 328: @phsym Thanks, wow how did I miss this 🤦

Issue #336, opened 2020-05-08
SetFormData() could support `mutlipart:form-data` by offloading to SetMultipartField?


ArcticSnowman:
I've been trying to make a PUT request to an API that wants the form data as `multipart:form-data`.  Finally worked out (after digging into the code) that SetMultipartField can be used if you don't set the filename and pass in the values via `strings.NewReader(value)`.

```
client.R().
SetMultipartField("branch", "", "", strings.NewReader(branch)).
SetMultipartField("sourceBranch", "", "", strings.NewReader(sourceBranch)).
SetMultipartField("content", "", "", strings.NewReader(mergeLines(content))).
```

It got me thinking that SetFormData could be enhanced or a new SetMultipartFormData() added. Such that it you can simply pass in a `map(string)string` of key/values paris and it handles the `strings.NewReader` bits to pass to `SetMultipartField()`

jeevatkm:
PR #337 is merged.

PR #337, opened 2020-05-09
Add SetMultipartFormData func to Request

ArcticSnowman:
Suggestion for #336

### DIFF
--- a/request.go
+++ b/request.go
@@ -322,6 +322,16 @@ func (r *Request) SetFileReader(param, fileName string, reader io.Reader) *Reque
 	return r
 }
 
+// SetMultipartFormData method allows simple form data to be attached to the request as `multipart:form-data`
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -322,6 +322,16 @@ func (r *Request) SetFileReader(param, fileName string, reader io.Reader) *Reque
 	return r
 }
 
+// SetMultipartFormData method allows simple form data to be attached to the request as `multipart:form-data`
+func (r *Request) SetMultipartFormData(data map[string]string) *Request {
```
jeevatkm, line 326: @ArcticSnowman Technically, the method `r.SetMultipartField` could be used directly for the described use case on #336. So the new method `SetMultipartFormData` is decorated/helper.

ArcticSnowman: Yeah it's totally a helper function just as a convenience

PR #341, opened 2020-05-12
add types for middleware

mloberg:
While working with middleware functions, it would be nice to reference a type rather than having to copy/paste the whole function.

Thanks for the work on Resty!

### DIFF
--- a/client.go
+++ b/client.go
@@ -66,6 +66,23 @@ var (
 	bufPool           = &sync.Pool{New: func() interface{} { return &bytes.Buffer{} }}
 )
 
+type (
+	// RequestMiddleware type is for request middleware, called before a request is sent
+	RequestMiddleware func(*Client, *Request) error
+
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -66,6 +66,15 @@ var (
 	bufPool           = &sync.Pool{New: func() interface{} { return &bytes.Buffer{} }}
 )
 
+type (
```
jeevatkm, line 69: @mloberg Can you please add godoc for the exported types?

mloberg: added

client.go
```diff
@@ -66,6 +66,23 @@ var (
...
+	// RequestMiddleware type is for request middleware, called before a request is sent
+	RequestMiddleware func(*Client, *Request) error
+
+	// ResponseMiddleware type is for response middleware, called after a response has been recieved
```
moorereason, line 73: s/recieved/received/

mloberg: :derp: fixed

PR #347, opened 2020-06-08
Fix trace timing for reused connections

### LINKED ISSUE #346: Getting improper value for TotalTime
Hi,

Wrote this simple program to check the effect of the total time when using a connection from a pool.

```
package main

import (
	"fmt"
)

func main() {
	httpClient := resty.New()

	for i := 0; i < 5; i++ {
		url := "http://httpbin.org/get"
		resp, _ := httpClient.R().EnableTrace().Get(url)

		ti := resp.Request.TraceInfo()

		fmt.Println("Loop count", i+1)
		fmt.Println("TotalTime    :", ti.TotalTime)
	}
}
```

In the first loop, the print for TotalTime is proper, however in the subsequent loops, the TotalTime is the same huge value as below:

Loop count 1
TotalTime    : 429.728934ms
Loop count 2
TotalTime    : **2562047h47m16.854775807s**
Loop count 3
TotalTime    : **2562047h47m16.854775807s**
Loop count 4
TotalTime    : **2562047h47m16.854775807s**
Loop count 5
TotalTime    : **2562047h47m16.854775807s**

Could someone please let me know if there is issue in the code above, or if it is caused by a bug in the library.

Thanks.

moorereason:
When a connection is reused, the normal trace hooks are skipped, which
led to an incorrect calculation of TotalTime (essentially doing a zero
minus now calculation). Detect connection reuse and use a different
calculation for TotalTime.

Fixes #346

### DIFF
--- a/request.go
+++ b/request.go
@@ -578,10 +578,15 @@ func (r *Request) TraceInfo() TraceInfo {
 		DNSLookup:     ct.dnsDone.Sub(ct.dnsStart),
 		TLSHandshake:  ct.tlsHandshakeDone.Sub(ct.tlsHandshakeStart),
 		ServerTime:    ct.gotFirstResponseByte.Sub(ct.gotConn),
-		TotalTime:     ct.endTime.Sub(ct.dnsStart),
 		IsConnReused:  ct.gotConnInfo.Reused,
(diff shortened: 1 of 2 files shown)


jeevatkm:
Travis Build: https://travis-ci.org/github/go-resty/resty/builds/696165904

PR #354, opened 2020-07-14
README: Add clarification on Request.RawRequest

lggomez:
This PR adds an additional bullet about `Request.RawRequest`, since consumers may wrongly assume that the native request is available at all times but in fact it is only instantiated once the resty Request starts executing an action


jeevatkm:
@lggomez Thanks for improving documentation.

PR #355, opened 2020-07-20
Fix - request: execute user defined middlewares after resty's internal middlewares

lggomez:
This allows middlewares to properly access the RawRequest instance and follow their expected behavior according to OnBeforeRequest:

```
// OnBeforeRequest method appends request middleware into the before request chain.
// Its gets applied after default Resty request middlewares and before request
// been sent from Resty to host server.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: go
 
-sudo: false
+os: linux
 
 go: # use travis ci resource effectively, keep always latest 2 versions and tip :) 
+  - 1.15.x
   - 1.14.x
-  - 1.13.x
   - tip
 
 install:
@@ -16,6 +16,6 @@ script:
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 
-matrix:
+jobs:
   allow_failures:
     - go: tip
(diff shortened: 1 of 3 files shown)


lggomez:
@jeevatkm it seems travis is stuck, the local tests run fine

If you're ok with this changeset post ci, would you mind making a v2.3.1 release so I can update from latest on my project? 

Thanks in advance

jeevatkm:
@lggomez Thank you for your PR. Seems like something is wrong in the travis integration. Will check it.

jeevatkm (approved):
@lggomez Thanks for the PR. I'm sorry for the delayed response. 

Also, Travis build has been addressed.

almas1992:
Why this merged change not included in V2.6.0 ？I would get raw request in custom middlewares , but the raw request is nill now .

ForsakenHarmony:
It was reverted https://github.com/go-resty/resty/pull/414

Is there any chance you'll add something like this back? @jeevatkm

NeoCN:
@jeevatkm Is there any release plan about this change? We can not upgrade to new versions without this.

jeevatkm:
@NeoCN Actually this PR was not a fix. That's why it was got reverted. I accidentally merged without realizing it on my bad day.
Resty middleware flow is designed this way from day one. If you would like to manipulate the RawRequest before sending out the request, kindly make use of PreRequestHook.

PR #362, opened 2020-08-03
Preserve marshal errors in middleware request body handler

### LINKED ISSUE #361: Missing error message when Marshal failed
If it fails during Marshal JSON or Marshal XML, the Marshal error message is lost. only get "unsupported 'Body' type/value" error. Can it return errors with the original error message?

master/middleware.go:477 
```
	if bodyBytes == nil && r.bodyBuf == nil {
		err = errors.New("unsupported 'Body' type/value")
	}

	// if any errors during body bytes handling, return it
	if err != nil {
		return
	}
```

moorereason:
Fixes #361


jeevatkm (approved):
@moorereason Thanks for taking care and PR.

PR #368, opened 2020-08-24
fix request setContentLength

lewisay:


jeevatkm (approved):
@lewisay Thanks for fixing the bug.

Issue #329, opened 2020-04-10
problem with  resty.Unmarshalc(..)


latenite4:
running on ubuntu 18.04 sending GET request to rapidAPI "coronavirus/cases_by_country.php" API.
API server responds with correct JSON and the Unmarshalc routine returns err == nil.
here is partial example of JSON response:
{
  "countries_stat": [
    {
      "country_name": "",
      "cases": "721",
      "deaths": "13",
      "region": "",
      "total_recovered": "619",
      "new_deaths": "0",
      "new_cases": "0",
      "serious_critical": "10",
      "active_cases": "89",
      "total_cases_per_1m_population": "0"
    },
    {
      "country_name": "USA",
      "cases": "466,969",
      "deaths": "16,636",
      "region": "",
      "total_recovered": "25,316",
      "new_deaths": "1,845",
      "new_cases": "31,939",
      "serious_critical": "10,011",
      "active_cases": "425,017",
      "total_cases_per_1m_population": "1,411"
    },

i have tried unmarshaling into 
type AutoGenerated2 struct {
	CountriesStat []struct {
		CountryName               string `json:"country_name"`
		Cases                     string `json:"cases"`
		Deaths                    string `json:"deaths"`
		Region                    string `json:"region"`
		TotalRecovered            string `json:"total_recovered"`
		NewDeaths                 string `json:"new_deaths"`
		NewCases                  string `json:"new_cases"`
		SeriousCritical           string `json:"serious_critical"`
		ActiveCases               string `json:"active_cases"`
		TotalCasesPer1MPopulation string `json:"total_cases_per_1m_population"`
	} `json:"countries_stat"`
	StatisticTakenAt string `json:"statistic_taken_at"`
} and also into
type AutoGenerated2 struct {
	CountriesStat [212]struct {
		CountryName               string `json:"country_name"`
		Cases                     string `json:"cases"`
		Deaths                    string `json:"deaths"`
		Region                    string `json:"region"`
		TotalRecovered            string `json:"total_recovered"`
		NewDeaths                 string `json:"new_deaths"`
		NewCases                  string `json:"new_cases"`
		SeriousCritical           string `json:"serious_critical"`
		ActiveCases               string `json:"active_cases"`
		TotalCasesPer1MPopulation string `json:"total_cases_per_1m_population"`
	} `json:"countries_stat"`
	StatisticTakenAt string `json:"statistic_taken_at"`
}
[...]

PR #371, opened 2020-09-04
add ForceContentType example

mattfenwick:
This is a really useful method for handling some odd cases, would be great to have an example in the README to make it easier to find!

Based on [this issue](https://github.com/go-resty/resty/issues/329#issuecomment-686957676).


jeevatkm (approved):
Thanks @mattfenwick for the PR to add documentation.

Issue #356, opened 2020-07-21
Allow OnAfterResponse ResponseMiddleware to run on all paths


lggomez:
As of now, custom response middlewares are only executed on the happy path of a request (that is, no error and with `notParseResponse == false` due to them being in the same list as the internal Resty middlewares for response handling

It is desirable that the user defined middlewares are executed on all paths and let the response handling to the user (as user defined request middlewares work now)

zhong-wx:
Due to middlewares cannot run on all paths, I have to design a new Middleware layer base on Wrapping Transport. QAQ

jeevatkm:
@lggomez I believe you're referring to response middlewares correct? If so what is the behavior you're trying to propose, do you mind showcasing via PR?

lggomez:
The work involved is similar to #355, plus a separation of `afterResponse` and `udAfterResponse` so that OnAfterRequest middlewares behave exactly that the OnBefore ones, internal ones first and client defined after

Now that you solved the ci issue (thanks!) I was hoping to take a shot at this soon

jeevatkm:
@lggomez You're welcome, it was a config issue with Travis build.

Resty's middleware behavior (typically it should) will halt the execution at middleware which had an error regardless of internal or user-defined middleware.

#355 is different (thanks again for submitting PR). I'm trying to understand the design & approach of "run on all paths", could you please describe?

zhong-wx:
I think "run on all paths" meas that afterResponse middlewares will be execute whether the request and beforeRequest middlewares are done successfully or not

lggomez:
It is as @zhong-wx says, I layed out the initial implementation on the linked PR

jeevatkm:
There is a middleware redesign coming in v3, so Resty users get to choose the middleware execution and its order. Plus this issue suggestion I will consider applying this on default flow.

Issue #345, opened 2020-06-02
RFC: create client let us choice not much more middleware.


0x010:
```
func createClient(hc *http.Client) *Client {
	if hc.Transport == nil {
		hc.Transport = createTransport(nil)
	}

	c := &Client{ // not setting lang default values
		QueryParam:         url.Values{},
		FormData:           url.Values{},
		Header:             http.Header{},
		Cookies:            make([]*http.Cookie, 0),
		RetryWaitTime:      defaultWaitTime,
		RetryMaxWaitTime:   defaultMaxWaitTime,
		JSONMarshal:        json.Marshal,
		JSONUnmarshal:      json.Unmarshal,
		jsonEscapeHTML:     true,
		httpClient:         hc,
		debugBodySizeLimit: math.MaxInt32,
		pathParams:         make(map[string]string),
	}

	// Logger
	c.SetLogger(createLogger())

	// default before request middlewares
	c.beforeRequest = []RequestMiddleware{
		parseRequestURL,
		parseRequestHeader,
		parseRequestBody,
		createHTTPRequest,
		addCredentials,
	}

	// user defined request middlewares
	c.udBeforeRequest = []RequestMiddleware{}

	// default after response middlewares
	c.afterResponse = []ResponseMiddleware{
		responseLogger,
		parseResponseBody,
		saveResponseIntoFile,
	}

	return c
}

```

the current createClient seems abit heavy with much more Middleware. 

Please allow use choice which middleware can been used.

jeevatkm:
@0x010 Thanks for reaching out with your thoughts. I would like to understand the viewpoint of `bit heavy`. 

From the design of resty, every mentioned middleware is needed otherwise the resty library will not work :(  
The only thing which stands out/out of place is `saveResponseIntoFile`.  

May I know the issue you're facing?

PR #372, opened 2020-09-08
WIP: Add user defined OnBeforeRequest middlewares on all response paths, improve documentation

lggomez:
Work for #356

### DIFF
--- a/client.go
+++ b/client.go
@@ -125,6 +125,7 @@ type Client struct {
 	udBeforeRequest    []RequestMiddleware
 	preReqHook         PreRequestHook
 	afterResponse      []ResponseMiddleware
+	udAfterResponse    []ResponseMiddleware
 	requestLog         RequestLogCallback
 	responseLog        ResponseLogCallback
 }
@@ -364,6 +365,10 @@ func (c *Client) NewRequest() *Request {
 //
 //				return nil 	// if its success otherwise return error
 //			})
+//
+// Please note that reading the request body on a middleware will drain it,
+// making it unavailable on the following middlewares and at request time.
+// For more information see https://golang.org/pkg/net/http/httputil/#DumpRequest
 func (c *Client) OnBeforeRequest(m RequestMiddleware) *Client {
 	c.udBeforeRequest = append(c.udBeforeRequest, m)
 	return c
@@ -371,15 +376,20 @@ func (c *Client) OnBeforeRequest(m RequestMiddleware) *Client {
 
 // OnAfterResponse method appends response middleware into the after response chain.
 // Once we receive response from host server, default Resty response middleware
-// gets applied and then user assigened response middlewares applied.
+// gets applied and then user assigned response middlewares applied.
+// Note that these middleware will execute on all paths, including request failure
 // 		client.OnAfterResponse(func(c *resty.Client, r *resty.Response) error {
 //				// Now you have access to Client and Response instance
 //				// manipulate it as per your need
 //
 //				return nil 	// if its success otherwise return error
 //			})
+//
+// Please note that reading the response body on a middleware will drain it,
+// making it unavailable on the following middlewares and at further consumption.
+// For more information see https://golang.org/pkg/net/http/httputil/#DumpResponse
 func (c *Client) OnAfterResponse(m ResponseMiddleware) *Client {
-	c.afterResponse = append(c.afterResponse, m)
+	c.udAfterResponse = append(c.udAfterResponse, m)
 	return c
 }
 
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -874,6 +893,15 @@ func (c *Client) execute(req *Request) (*Response, error) {
 	return response, wrapNoRetryErr(err)
 }
 
+// Apply user defined Response middlewares on the given response
+func (c *Client) applyCustomResponseMiddlewares(response *Response) {
```
jeevatkm, line 897: @lggomez `applyCustomResponseMiddlewares` => `applyUdResponseMiddlewares` please.

client.go
```diff
@@ -125,6 +125,7 @@ type Client struct {
 	udBeforeRequest    []RequestMiddleware
 	preReqHook         PreRequestHook
 	afterResponse      []ResponseMiddleware
+	udAfterResponse    []ResponseMiddleware
```
jeevatkm, line 128: @lggomez `udAfterResponse` initialize it when a resty client gets created.
Like -
https://github.com/go-resty/resty/blob/d2ba00f5a573e785d8d90e05433d3f4ef4ad3665/client.go#L969

jeevatkm (changes requested):
@lggomez Thanks for creating PR to describing an enhancement. Now, I have an understanding of the proposal.
* I have shared few comments, please take care.
* Any reason, why these two (`responseLogger`, `parseResponseBody`) response middleware is not called in the proposal?  At least I think logger one needs to be executed.

Thanks again!

lggomez:
I made the suggested changes

> Any reason, why these two (responseLogger, parseResponseBody) response middleware is not called in the proposal? At least I think logger one needs to be executed.

I would like to be able to use a custom logger (at least in my use case), and I can see the default `responseLogger` mw injected which is debug only as a basis for #345. `parseResponseBody`, at least, can be disabled via `SetDoNotParseResponse`, but I can't see a similar experience for this one yet (please correct me if I'm wrong)

For the general case of intercepting all return paths (including errors where responses can be in an invalid state) I'd leave that responsibility up to the user and their middleware implementation (unless otherwise specified in the documentation like the body drain mentions)

lggomez:
To clarify a bit more about the use case: my application is designed to support debug flows on a per-request basis based on a flag so the current per-client configuration doesn't really fit and a user defined middleware in this setup could allow for per-request behaviors without coupling to the internal middlewares

jeevatkm:
@lggomez I see your viewpoint and use case. Let's do the following -
* With this PR, let's stick to current behavior, .i.e execute `responseLogger`, `parseResponseBody` middleware for all paths along with user-defined ones.
* As part #345 solution exposing all middlewares to resty users via new sub-package 🤔 . This is maybe a breaking change, need to analyze a bit.
* Also bringing debug setting per request is also an option.

Sounds like a plan?

lggomez:
All of them sound fine; for the scope of the PR I'd keep with implementing just the first item for now

Regarding that, one of the issues of moving the internal middleware to all paths is that suddenly the applyResponseMiddlewares has to return an error:

- For error paths, we don't use the error returned by it (it will break behavior and tests otherwise)
- For the normal return path, use its error result for the `wrapNoRetryErr(err)` return value

I pushed the current change, so I'll await your opinion on this

jeevatkm:
> For error paths, we don't use the error returned by it (it will break behavior and tests otherwise)

@lggomez Do you mean returning an error of response middleware from L852, L866, and L875 it will break the existing behavior? 

If it is not returned; then the previous error is getting eaten/suppressed by resty, isn't it?. Typically any libraries shouldn't eat any errors, only the end-user can make a decision on ignoring the error.

lggomez:
Yeah, altough it seems to happen only on the first path , changing
```go
                c.applyResponseMiddlewares(response)
		return response, err
```

to


```go
		return response, c.applyResponseMiddlewares(response)
```

breaks the following (only on the first early return on the execute() method):

```
--- FAIL: TestClientRedirectPolicy (0.03s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x18 pc=0x77746d]

goroutine 84 [running]:
testing.tRunner.func1.1(0x815340, 0xbe0890)
	C:/Go/src/testing/testing.go:988 +0x314
testing.tRunner.func1(0xc0001e8a20)
	C:/Go/src/testing/testing.go:991 +0x400
panic(0x815340, 0xbe0890)
	C:/Go/src/runtime/panic.go:969 +0x174
github.com/go-resty/resty/v2.TestClientRedirectPolicy(0xc0001e8a20)
	C:/Users/a309586/DEV/GITHUB/resty/client_test.go:110 +0x2dd
testing.tRunner(0xc0001e8a20, 0x8a7428)
	C:/Go/src/testing/testing.go:1039 +0xe3
created by testing.(*T).Run
	C:/Go/src/testing/testing.go:1090 +0x379
```

This issue aside, the applyResponseMiddlewares() error would be shadowing the original execute() error, so doing this right is a bit hairier than I thought initially. We want to be able to cancel the middleware stack but we also want to be able to propagate accordingly the errors (maybe wrap the mw error?)

jeevatkm:
@lggomez I'm sorry for the delay in response.

I think wrapping the error will be a good choice. Please go ahead. Thanks.  

Also, streamlining this part in the next major version as breaking change will help everyone.

PR #374, opened 2020-09-09
Couple of enhancements

### LINKED ISSUE #352: Find a way to fetch the number of attempt
Retry comes with a nice retry system.
For monitoring reason, I would like to be able when I get a response, how many attempts Resty has to do to complete the request.

Something like that:
```
client := resty.New()SetRetryCount(3).OnAfterResponse(func(c *resty.Client, resp *resty.Response) error {
			metrics.ReportHTTPCall(resp.StatusCode(),resp.Request.Method, somewhere.NumberOfAttempt())
			return nil
		})

```

Is there a way to get that value?

### LINKED ISSUE #370: collect more http trace info like remote IP which is a import message
to do that, we just add more field in TraceInfo struct
and remote IP can get from Request.clientTrace.ci.gotConnInfo.Conn.RemoteAddr()

jeevatkm:
Resolves #352 
Resolves #370

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
   * v2.3.0 [released](https://github.com/go-resty/resty/releases/tag/v2.3.0) and tagged on May 20, 2020.
   * v2.0.0 [released](https://github.com/go-resty/resty/releases/tag/v2.0.0) and tagged on Jul 16, 2019.
-  * v1.12.0 [released](https://github.com/go-resty/resty/releases/tag/v1.12.0) and tagged on Feb 27, 2019.  
+  * v1.12.0 [released](https://github.com/go-resty/resty/releases/tag/v1.12.0) and tagged on Feb 27, 2019.
   * v1.0 released and tagged on Sep 25, 2017. - Resty's first version was released on Sep 15, 2015 then it grew gradually as a very handy and helpful library. Its been a two years since first release. I'm very thankful to Resty users and its [contributors](https://github.com/go-resty/resty/graphs/contributors).
 
 ## Features
@@ -55,11 +55,12 @@
   * Option to specify expected `Content-Type` when response `Content-Type` header missing. Refer to [#92](https://github.com/go-resty/resty/issues/92)
   * Resty design
     * Have client level settings & options and also override at Request level if you want to
-    * Request and Response middlewares
+    * Request and Response middleware
     * Create Multiple clients if you want to `resty.New()`
     * Supports `http.RoundTripper` implementation, see [SetTransport](https://godoc.org/github.com/go-resty/resty#Client.SetTransport)
     * goroutine concurrent safe
     * Resty Client trace, see [Client.EnableTrace](https://godoc.org/github.com/go-resty/resty#Client.EnableTrace) and [Request.EnableTrace](https://godoc.org/github.com/go-resty/resty#Request.EnableTrace)
+      * Since v2.4.0, `RequestAttempt` value supported in trace info also Request instance contains `Attempt` attribute
     * Debug mode - clean and informative logging presentation
     * Gzip - Go does it automatically also resty has fallback handling too
     * Works fine with `HTTP/2` and `HTTP/1.1`
@@ -83,7 +84,7 @@
 
 #### Supported Go Versions
 
-Initially Resty started supporting `go modules` since `v1.10.0` release. 
+Initially Resty started supporting `go modules` since `v1.10.0` release.
 
 Starting Resty v2 and higher versions, it fully embraces [go modules](https://github.com/golang/go/wiki/Modules) package release. It requires a Go version capable of understanding `/vN` suffixed imports:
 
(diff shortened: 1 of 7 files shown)


moorereason (changes requested):
I don't have permission to push to your PR branch, so here's a patch to test the RemoteAddr feature:

```patch
diff --git a/request_test.go b/request_test.go
index b02e04c..1035d87 100644
--- a/request_test.go
+++ b/request_test.go
@@ -1644,6 +1644,8 @@ func TestTraceInfo(t *testing.T) {
        ts := createGetServer(t)
        defer ts.Close()

+       serverAddr := ts.URL[strings.LastIndex(ts.URL, "/")+1:]
+
        client := dc()
        client.SetHostURL(ts.URL).EnableTrace()
        for _, u := range []string{"/", "/json", "/long-text", "/long-json"} {
@@ -1660,6 +1662,7 @@ func TestTraceInfo(t *testing.T) {
                assertEqual(t, true, tr.TotalTime >= 0)
                assertEqual(t, true, tr.TotalTime < time.Hour)
                assertEqual(t, true, tr.TotalTime == resp.Time())
+               assertEqual(t, tr.RemoteAddr.String(), serverAddr)
        }

        client.DisableTrace()
```

jeevatkm:
Sure, I will add it to the test case.

> I don't have permission to push to your PR branch

Hmm, this is strange. You're a member of this repo, you should have write access. PR branch resides in this repo. Can you share the error message?

jeevatkm:
@moorereason Test case updated

moorereason:
I see that I'm a "member" of the project, but I appear to only have read access.
```
sh$ git config --get remote.origin.url
git@github.com:go-resty/resty.git

sh$ git status
On branch few-enhancements
Your branch is ahead of 'origin/few-enhancements' by 1 commit.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean

sh$ git push
ERROR: Permission to go-resty/resty.git denied to moorereason.`
fatal: Could not read from remote repository.

**Comment by jeevatkm** [comment]
@moorereason It's my bad, I created @go-resty/core team but didn't assign it to the repo. I'm sorry. Can you try now?

Also, please review this PR and approve it.

**Comment by moorereason** [comment]
First, I don't think `RetryAttempt` needs to be added to the `TraceInfo` object.  Simply exporting it on `Request` is enough.

Second, can you update the README.md to include `RemoteAddr` in the `TraceInfo` example?

Finally, wouldn't it be helpful to set the `RetryAttempt` for each retry request so that the OnAfter/middleware funcs have access to it?  For example...

Given this test program:

```go
package main

import (
        "fmt"
        "net/http"

)

func main() {
        client := resty.New().
                EnableTrace().
                SetRetryCount(3).
                AddRetryCondition(func(r *resty.Response, e error) bool {
                        return r.StatusCode() == http.StatusNotFound
                }).
                OnAfterResponse(func(c *resty.Client, resp *resty.Response) error {
                        fmt.Printf("OnAfterResponse:attempt = %d\n", resp.Request.RetryAttempt)
                        return nil
                })

        resp, _ := client.R().Get("https://httpbin.org/status/404")
        fmt.Printf("%d\n", resp.Request.RetryAttempt)
}
```
The current PR executes as:
```
sh$ go run main.go
OnAfterResponse:attempt = 0
OnAfterResponse:attempt = 0
OnAfterResponse:attempt = 0
OnAfterResponse:attempt = 0
4
```

With this change:
```patch
diff --git a/request.go b/request.go
index 9e49fca..51f0520 100644
--- a/request.go
+++ b/request.go
@@ -699,6 +699,7 @@ func (r *Request) Execute(method, url string) (*Response, error) {
                        attempt++

                        r.URL = r.selectAddr(addrs, url, attempt)
+                       r.RetryAttempt = attempt

                        resp, err = r.client.execute(r)
                        if err != nil {
```

The results become:
```
sh$ go run main.go
OnAfterResponse:attempt = 1
OnAfterResponse:attempt = 2
OnAfterResponse:attempt = 3
OnAfterResponse:attempt = 4
4
```

moorereason:
It looks like the permissions are fixed now.  Thanks 👍

jeevatkm:
@moorereason why do you feel `RetryAttempt` value is not a candidate for TraceInfo?  

Sure, I will update the code and readme for remaining.

moorereason:
> why do you feel `RetryAttempt` value is not a candidate for TraceInfo?

Because it's already exposed in the `Request` object.  Why copy the value to `Request.TraceInfo` when it's already on the parent object?  Maybe I'm misunderstanding how people are using `TraceInfo`.  Why do you want to add it to `TraceInfo`?

jeevatkm:
Typically, trace info or metric values are serialized as JSON sent to metric capture systems like influxdb, prometheus, datadog, newrelic, etc. Currently, resty users are doing this step at their end.

I have been thinking for a long; still, I have not decided field name format snake case vs camel case or support both via function on `r.TraceInfo().ToJSON(resty.FieldCamelCase)`.  That's why no traction on this.

moorereason:
Thanks for the explanation.  Makes sense.  :+1:

I have a couple more comments:

1. We should update README.md to include info about `RetryAttempt` on the `Request` and `TraceInfo` objects.

1. I'm not sure that `RetryAttempt` is a good property name.  If no retry scenario occurred (the request happened the first time without issue), `RetryAttempt` would be equal to `1`.  In my mind, there were no "retry attempts" in that scenario.  It would be more descriptive to name the property `ConnectionAttempt` or `ConnAttempt` or something.  What do you think?

jeevatkm:
Thanks @moorereason 

> 1. We should update README.md to include info about `RetryAttempt` on the `Request` and `TraceInfo` objects.

Sure, I will add details in the readme 👍.

> 2. I'm not sure that `RetryAttempt` is a good property name.  If no retry scenario occurred (the request happened the first time without issue), `RetryAttempt` would be equal to `1`.  In my mind, there were no "retry attempts" in that scenario.  It would be more descriptive to name the property `ConnectionAttempt` or `ConnAttempt` or something.  What do you think?

Nice, good one. The `Connection or Conn` word might collide with TCP connection info. How about this?
* In the Request struct let's call it as `Attempt` -> it will bring the meaning of `Request.Attempt`
* In the TraceInfo struct let's call it as `RequestAttempt` -> it will bring the meaning of `TraceInfo().RequestAttempt`

After your response, I will prepare this PR as final.

moorereason:
> How about this?
> 
> * In the Request struct let's call it as `Attempt` -> it will bring the meaning of `Request.Attempt`
> * In the TraceInfo struct let's call it as `RequestAttempt` -> it will bring the meaning of `TraceInfo().RequestAttempt`

Sounds good.  :+1:

jeevatkm:
@moorereason I have updated it. This PR is ready for final review.

README.md
```diff
@@ -55,11 +55,12 @@
   * Option to specify expected `Content-Type` when response `Content-Type` header missing. Refer to [#92](https://github.com/go-resty/resty/issues/92)
   * Resty design
     * Have client level settings & options and also override at Request level if you want to
-    * Request and Response middlewares
+    * Request and Response middleware
     * Create Multiple clients if you want to `resty.New()`
     * Supports `http.RoundTripper` implementation, see [SetTransport](https://godoc.org/github.com/go-resty/resty#Client.SetTransport)
     * goroutine concurrent safe
     * Resty Client trace, see [Client.EnableTrace](https://godoc.org/github.com/go-resty/resty#Client.EnableTrace) and [Request.EnableTrace](https://godoc.org/github.com/go-resty/resty#Request.EnableTrace)
+      * Since v2.4.0, `RequestAttempt` value supported in trace info also Request instance contains `Attempt` attribute
```
moorereason, line 63: Rewording suggestion:

Since v2.4.0, trace info contains a `RequestAttempt` value, and the Request object contains an `Attempt` attribute

moorereason (approved):
Small rewording suggestion, but I won't hold up the PR for it.  Nice work. 👍

README.md
jeevatkm: Thanks, @moorereason, I will update the readme with suggestion on the next round. I have to do a few other things.

Issue #300, opened 2020-01-24
Getting Body size 0


grafael:
For some **POST** requests I'm getting Body length = 0 error. 
I did the exact same request using curl and worked properly, also I did use pure **http.request** and worked. I have no idea what else I could configure to get that working.

Snippet:

```golang
data := `{"name":"test"}`
resp, err := c.client.New().R().
	SetHeader("Content-Type", "application/json").
	SetBody(data).
	Post(url)
```
 
http: ContentLength=181 with Body length 0

```
Error      : Post <my_endpoint_here>: http: ContentLength=181 with Body length 0
Status Code: 0
Status     :
Time       : 175.825µs
Received At: 2020-01-24 08:56:39.710017 -0300 -03 m=+301.171273859
Body       :

Params     :
 map[]

Request Trace Info:
DNSLookup    : 31.906079ms
ConnTime     : 6.764µs
TLSHandshake : 335.204955ms
ServerTime   : -189.941µs
ResponseTime : 319.94µs
TotalTime    : 175.825µs
IsConnReused : true
IsConnWasIdle: true
ConnIdleTime : 96.155µs
```

jeevatkm:
@grafael Would you mind providing an example program to demo the issue?

jeevatkm:
@grafael Any updates? I think you could also explore the resty test cases to validate the resty.

wjh000123:
@jeevatkm this issue occurred when the reponse is in HTTP status - 307.

manojkva:
I had this same issue while using resty v2.2.0 with Gin framework. Resty was adding up an additional "/" at the end of the Endpoint even when not provided. 
This was not allowed in Gin i suppose (https://github.com/gin-gonic/gin/issues/1004).
I got it resolved by downgrading my resty version to v2.0.0 .
Please note that v2.1.0 too had the same issue.

jeevatkm:
@wjh000123  @manojkva Thanks for the inputs, will have a look.

alexio777:
I have exactly the same problem.

`	resp, err := hc.rc.R().
		SetHeader("Content-Type", "application/octet-stream").
		SetBody(data).
		Post(fmt.Sprintf("%s%s", gRemoteAPIURL, path))
`
where data is []byte

and I get an error: http: ContentLength=224 with Body length 0

moorereason:
@alexio777, can you provide a complete test program to demonstrates the issue?

jeevatkm:
[...]

PR #380, opened 2020-09-27
fix: get body size 0 (#300)

shiguangyin:
when the http status code is 307 or 308, the request.Body will be closed, after that the func GetBody will be called but the bodyBuf was closed, then we will get an error http: ContentLength=xxx with Body length 0

### DIFF
--- a/client.go
+++ b/client.go
@@ -54,6 +54,7 @@ var (
 	hdrContentLengthKey   = http.CanonicalHeaderKey("Content-Length")
 	hdrContentEncodingKey = http.CanonicalHeaderKey("Content-Encoding")
 	hdrAuthorizationKey   = http.CanonicalHeaderKey("Authorization")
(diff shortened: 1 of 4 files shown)


shiguangyin:
see https://github.com/golang/go/blob/72a9dec156408a87548deb920a67b8bf787062db/src/net/http/client.go#L737

jeevatkm (changes requested):
@shiguangyin Thanks for the PR to fix the body size issue on redirects.

Can you please improve unit test cases for missed lines? you can see the code coverage report here https://github.com/go-resty/resty/pull/380#issuecomment-699653309

shiguangyin:
@jeevatkm I think the code coverage report is not correct, I don't know why the code coverage decreased, I did't have any change on some files like redirect.go and util.go, but the report show those files are impacted. I don't  know how to improve the unit test cases to increase the code coverage

PR #384, opened 2020-10-30
Added AddRetryAfterErrorCondition() option

gtpaulose:
A really simple option so folks can set the retry options with easy and less verbosity in their code.

### DIFF
--- a/client.go
+++ b/client.go
@@ -554,6 +554,19 @@ func (c *Client) AddRetryCondition(condition RetryConditionFunc) *Client {
 	return c
 }
 
+// AddRetryAfterErrorCondition adds the basic condition of retrying after enncoutering
+// an error from the http response
+func (c *Client) AddRetryAfterErrorCondition() *Client {
+	c.AddRetryCondition(func(response *Response, err error) bool {
+		if response.IsError() {
+			return true
+		}
+
+		return false
+	})
+	return c
+}
+
 // SetTLSClientConfig method sets TLSClientConfig for underling client Transport.
 //
 // For Example:
--- a/client.go
+++ b/client.go
@@ -554,7 +554,7 @@ func (c *Client) AddRetryCondition(condition RetryConditionFunc) *Client {
 	return c
 }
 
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -554,6 +554,19 @@ func (c *Client) AddRetryCondition(condition RetryConditionFunc) *Client {
 	return c
 }
 
+// AddRetryAfterErrorCondition adds the basic condition of retrying after enncoutering
```
jeevatkm, line 557: @gtpaulose PR looks good, can you please correct typo `enncoutering` -> `encountering`?

client.go
```diff
@@ -554,6 +554,19 @@ func (c *Client) AddRetryCondition(condition RetryConditionFunc) *Client {
...
+// an error from the http response
+func (c *Client) AddRetryAfterErrorCondition() *Client {
+	c.AddRetryCondition(func(response *Response, err error) bool {
+		if response.IsError() {
```
jeevatkm, line 561: @gtpaulose can you please add a test case for success and failure flow of new condition?
More info: https://codecov.io/gh/go-resty/resty/pull/384/diff?src=pr&el=tree#diff-Y2xpZW50Lmdv

jeevatkm (changes requested):
@gtpaulose Thanks for the PR, surely this handy for the resty users.

Please take a look at a couple of comments.

client.go
gtpaulose: Hey did you mean the test case for 'refiring under an error and non error scenario (i.e `response.IsError() == true` and `response.IsError() == false` )? '

moorereason: Yes

gtpaulose: Alright, ive added the test case. Ive reused an existing path in the test get server switch case, hope thats okay!

gtpaulose:
Is there anymore changes required?

jeevatkm (approved):
@gtpaulose Thanks for the updates and for taking care of the review comments. 
I'm really sorry for the delay, due to personal reasons.

PR #392, opened 2020-11-30
fix:when SetRetryCount -1, both resp and error are nil.

yjhmelody:
I add a test for this behavior, I think it is a bug.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -590,6 +590,7 @@ func TestLogCallbacks(t *testing.T) {
 		Get(ts.URL + "/profile")
 	assertEqual(t, errors.New("request test error"), err)
 	assertNil(t, resp)
+	assertNotNil(t, err)
 
 	c.OnRequestLog(nil)
 	c.OnResponseLog(func(r *ResponseLog) error { return errors.New("response test error") })
--- a/request_test.go
+++ b/request_test.go
@@ -48,6 +48,16 @@ func TestGet(t *testing.T) {
 	logResponse(t, resp)
 }
 
+func TestIllegalRetryCount(t *testing.T) {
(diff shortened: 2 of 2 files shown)


jeevatkm (commented):
@yjhmelody Thanks for creating a PR. I'm sorry I'm not able to grasp the outcome.

Can you explain and help me to understand?

yjhmelody:
@jeevatkm 

https://github.com/go-resty/resty/blob/master/retry.go#L87

```go
for attempt := 0; attempt <= opts.maxRetries; attempt++ {
...
```
opts.MaxRetries is `-1` so we cannot go into the loop,
The `Backoff`'s all options have not done anything exactly.

client_test.go
```diff
@@ -612,118 +612,3 @@ func TestNewWithLocalAddr(t *testing.T) {
 	assertNil(t, err)
 	assertEqual(t, resp.String(), "TestGet: text response")
 }
-
-func TestClientOnResponseError(t *testing.T) {
```
moorereason, line 616: Why are these two test functions removed?

yjhmelody: I'm sorry that I do a wrong rebased operation. I will fix it.

jeevatkm:
@yjhmelody This PR contains new test cases added. But you mentioned it as "fix". What is the bug and fix you're submitting?

moorereason:
I created issue #404 to track the actual issue reported in the PR.

yjhmelody:
@moorereason Oh, let we choose the second solution ?

jeevatkm (approved):
@yjhmelody Thanks for your contributions. I'm really sorry for the delay, due to personal reasons.

PR #398, opened 2020-12-22
feature: add OnError hook

justenwalker:
Adds a new hook that is called when a rest request returns an error.
This is called when the err returned by client.execute is non-nil, and
it is only called after all retry attempts have been exhausted.

This feature is intended to support logging, metrics, tracing on request
errors that otherwise cannot be captured with OnAfterResponse, such as
TLS Connection errors. It also provides a way to report response errors
only once after retries have been attempted, to avoid emitting error
logs when a request ultimately succeeds.

### DIFF
--- a/client.go
+++ b/client.go
@@ -388,7 +388,7 @@ func (c *Client) OnAfterResponse(m ResponseMiddleware) *Client {
 	return c
 }
 
-// OnError method adds a callback that will be run whenever a request execution fails
+// OnError method adds a callback that will be run whenever a request execution fails.
 // This is called after all retries have been attempted (if any).
 // If there was a response from the server, the error will be wrapped in *ResponseError
 // which has the last response received from the server.
@@ -922,8 +922,8 @@ func (c *Client) outputLogTo(w io.Writer) *Client {
 	return c
 }
 
-// ResponseError is a wrapper for including the server response with an error
-// neither the err nor the response should be nil
+// ResponseError is a wrapper for including the server response with an error.
+// Neither the err nor the response should be nil.
 type ResponseError struct {
 	Response *Response
 	Err      error
@@ -937,8 +937,8 @@ func (e *ResponseError) Unwrap() error {
 	return e.Err
 }
 
-// helper to run onErrorHooks hooks.
-// it wraps the error in a ResponseError if the resp is not nil
+// Helper to run onErrorHooks hooks.
+// It wraps the error in a ResponseError if the resp is not nil
 // so hooks can access it.
 func (c *Client) onErrorHooks(req *Request, resp *Response, err error) {
 	if err != nil {
--- a/client.go
+++ b/client.go
@@ -121,17 +121,17 @@ type Client struct {
 	debugBodySizeLimit int64
 	outputDirectory    string
 	scheme             string
-	pathParams      map[string]string
-	log             Logger
-	httpClient      *http.Client
-	proxyURL        *url.URL
-	beforeRequest   []RequestMiddleware
-	udBeforeRequest []RequestMiddleware
-	preReqHook      PreRequestHook
-	afterResponse   []ResponseMiddleware
-	requestLog      RequestLogCallback
-	responseLog     ResponseLogCallback
-	errorHooks      []ErrorHook
+	pathParams         map[string]string
+	log                Logger
+	httpClient         *http.Client
+	proxyURL           *url.URL
+	beforeRequest      []RequestMiddleware
+	udBeforeRequest    []RequestMiddleware
+	preReqHook         PreRequestHook
+	afterResponse      []ResponseMiddleware
+	requestLog         RequestLogCallback
+	responseLog        ResponseLogCallback
+	errorHooks         []ErrorHook
 }
 
 // User type is to hold an username and password information
@@ -390,10 +390,15 @@ func (c *Client) OnAfterResponse(m ResponseMiddleware) *Client {
 
 // OnError method adds a callback that will be run whenever a request execution fails
 // This is called after all retries have been attempted (if any).
-// 		client.OnError(func(err error) {
+// If there was a response from the server, the error will be wrapped in *ResponseError
+// which has the last response received from the server.
+//
+//		client.OnError(func(err error) {
+//			if v, ok := err.(*resty.ResponseError); ok {
+//				// Do something with v.Response
+//			}
 //			// Log the error, increment a metric, etc...
 //		})
-//
 func (c *Client) OnError(h ErrorHook) *Client {
 	c.errorHooks = append(c.errorHooks, h)
 	return c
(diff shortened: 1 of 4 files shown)


justenwalker:
This indirectly allows #356 since you can use OnAfterResponse to return an custom error that includes the response object, and therefore react to it from OnResponseError:

```go
type ResponseError struct {
	Response *resty.Response
}
func (r *ResponseError) Error() string {
	return r.Response.Status()
}

/// ...

	client.OnAfterResponse(func(client *resty.Client, response *resty.Response) error {
		if response.IsError() {
			return &ResponseError{Response: response}
		}
		return nil
	})
	client.OnResponseError(func(err error) {
		var re *ResponseError
		if errors.As(err, &re) {
			// Can use re.Response to get response body, status, etc...
		}
	})
```

justenwalker:
One other thing I'm thinking about is if `OnResponseError` is a good name for this hook, since a response is not in the hook, and a request may not have even been executed (in the case of a TLS error). Maybe this should be renamed to just `OnError`

justenwalker:
In the absence of any feedback, I've implemented the above enhancements:

1. Renamed OnResponseError to OnError (and related types/funcs)
2. The `err` received by the `ErrorHook` may now be wrapped in a `*ResponseError` which has the last `*Response` from the server. This can either be extracted using `errors.As` or a type assertion, so it should be backward compatible with all Go versions.

**Note**: I'm going the route of wrapping the error rather than changing the function signature so that it is more difficult for users of the library to unintentionally nil-dereference the response. Although the type assertion is arguably uglier to look at, it is more difficult to misuse, which I think is a good trade-off.

## type assertion

```go
client.OnError(func(err error) {
	if v, ok := err.(*resty.ResponseError); ok {
		resp := v.Response()
		// Do something with v.Response
	}
	// Log the error, increment a metric, etc...
})
```

## errors.As

```go
client.OnError(func(err error) {
	var v *resty.ResponseError
	if errors.As(err, &v) {
		// Do something with v.Response
	}
	// Log the error, increment a metric, etc...
})
```

justenwalker:
One other thing that might warrant a signature change for the hook is to make the OnError also include the `*Request`, since the request is always available.

```go
ErrorHook func(*Request, error)
```

client.go
```diff
@@ -384,6 +388,22 @@ func (c *Client) OnAfterResponse(m ResponseMiddleware) *Client {
 	return c
 }
 
+// OnError method adds a callback that will be run whenever a request execution fails
```
moorereason, line 391: End sentence with period.

client.go
```diff
@@ -902,6 +922,35 @@ func (c *Client) outputLogTo(w io.Writer) *Client {
 	return c
 }
 
+// ResponseError is a wrapper for including the server response with an error
```
moorereason, line 925: End sentence with period.  Start the next sentence with a capital letter.

client.go
```diff
@@ -902,6 +922,35 @@ func (c *Client) outputLogTo(w io.Writer) *Client {
...
+func (e *ResponseError) Error() string {
+	return e.Err.Error()
+}
+
+func (e *ResponseError) Unwrap() error {
+	return e.Err
+}
+
+// helper to run onErrorHooks hooks.
+// it wraps the error in a ResponseError if the resp is not nil
```
moorereason, line 941: Start sentence with capital letter.

client_test.go
```diff
@@ -612,3 +613,99 @@ func TestNewWithLocalAddr(t *testing.T) {
...
+				SetRetryCount(0).
+				SetRetryMaxWaitTime(time.Microsecond).
+				AddRetryCondition(func(response *Response, err error) bool {
+					if err != nil {
+						return true
+					}
+					return response.IsError()
+				}).
+				OnError(func(r *Request, err error) {
+					errs++
```
moorereason, line 687: Should access `r` in this test to ensure that we're properly passing the request into the hook.

Also, it doesn't look like you're testing multiple OnError hooks in a single request.  If we're going to keep that ability, we ought to test it.

moorereason (changes requested):
Requesting a few changes.

One question: why allow multiple `OnError` hooks to be defined?  What's the use case?

client_test.go
justenwalker: > One question: why allow multiple `OnError` hooks to be defined? What's the use case?

Don't have a particular use-case in mind for multiple on-error hooks, but it seemed like using a a range makes it at least possible for one to arise without a code change to accommodate it; and it doesn't make the implementation any less clean.

justenwalker:
> One question: why allow multiple `OnError` hooks to be defined? What's the use case?

Don't have a particular use-case in mind for multiple on-error hooks, but using a slice/range makes it at least possible for one to arise without a code change to accommodate it. It doesn't make the implementation any less clean. (ranging over nil is the same as ranging over an empty list, so we don't have to do any nil checks)

client.go
```diff
@@ -937,7 +937,7 @@ func (e *ResponseError) Unwrap() error {
 	return e.Err
 }
 
-// helper to run onErrorHooks hooks.
+// Helper to run onErrorHooks hooks.
 // it wraps the error in a ResponseError if the resp is not nil
```
moorereason, line 941: Capitalize "it" as well.

moorereason (approved):

moorereason:
Thanks for all the hard work and iterating/improving your idea.

I approved this change, but that doesn't mean it will be merged.  We'll wait for @jeevatkm to chime in.  I consider myself a "junior" member of this project, so I defer to @jeevatkm for all merges.

Also, if @jeevatkm approves of this PR, we'll want to update the README where appropriate.

jeevatkm (changes requested):
@justenwalker Thanks for your PR and iterative improvements while working on the PR.  Please go ahead and update the README then we merge it.

@moorereason Thanks for your review and help.

justenwalker:
> @justenwalker Thanks for your PR and iterative improvements while working on the PR. Please go ahead and update the README then we merge it.

README updated. Also noticed that it did not call the hook on SRV lookup failure, so I added that.

jeevatkm (approved):
@justenwalker Thanks for the README update. 

While looking at the SRV error update, I felt `OnError` could be improved with all errors in the Resty. That can be done later on.

Thanks, again 😄

justenwalker:
> While looking at the SRV error update, I felt `OnError` could be improved with all errors in the Resty. That can be done later on.

@jeevatkm, do you mean including the validation error? To be frank, I think that validation check might be more trouble than it's worth. There are a lot of so-called "REST" APIs out there that do not follow standards. For example, they accept a Body in a DELETE request, etc... If you want go-resty to work with them too, I'd suggest it removing that validation check entirely. 

But, that's outside the scope of this PR.

Anyway, Thanks for Merging!

jeevatkm:
> I think that validation check might be more trouble than it's worth. There are a lot of so-called "REST" APIs out there that do not follow standards.

@justenwalker I see, it makes sense. Thanks!

> I'd suggest it removing that validation check entirely.

@justenwalker I thought about it since the latest RFC getting an update for `GET` request body. I think it can be done in Resty v3.

PR #403, opened 2021-01-11
Enhancements

### LINKED ISSUE #387: Default Authorization Key name change
Hello.

I am in need of changing default Authorization key name (**hdrAuthorizationKey** var) to **"X-Authorization"**

https://thingsboard.io/docs/reference/rest-api/

* client:go
https://github.com/go-resty/resty/blob/master/client.go#L56
https://github.com/go-resty/resty/blob/master/client.go#L92

* middleware.go
https://github.com/go-resty/resty/blob/master/middleware.go#L252

My proposition is to add new exported element in **Client** struct + exported method **SetAuthorizationKeyName**. middleware.go then should check if value is set and make proper changes. 

**Edit:** I have just found **OnBeforeRequest** method: https://github.com/go-resty/resty/blob/master/client.go#L359
I will use that to add header, but it still would be nice to have an option I have wrote about ;)

Cheers, Paweł.

### LINKED ISSUE #395: How to set Content-Length header
I'm trying to call an API, that expects 'Content-Length' = 0 for a HTTP PATCH Request. There is no request body for this request. 

It seems golang (from 1.8) does not allow to set Content-length with zero bytes. ( pls see -- https://github.com/golang/go/issues/20257)

How do I set the header?
I have tried
``` 
restyClient.SetBody(http.NoBody).SetContentLength(true)
```

And
```
reqHeader := map[string]string {
		"Content-Length": "0",
}
```

None of these 2 ways set the content-length header.

### LINKED ISSUE #402: SetPathParam (add one parameter)
Hey there!
Thanks for supporting this project!

I'd like to add one small thing that will save some lines of code for me
I want to propose a function
```
func (r *Request) SetPathParam(p, v string) *Request {
    r.pathParams[p] = v
    return r
}
```

What do you think?

jeevatkm:
Closes #395 
Closes #387 
Closes #402

### DIFF
--- a/client.go
+++ b/client.go
@@ -742,6 +742,22 @@ func (c *Client) SetDoNotParseResponse(parse bool) *Client {
 	return c
 }
 
+// SetPathParam method sets single URL path key-value pair in the
+// Resty client instance.
+// 		client.SetPathParam("userId", "sample@sample.com")
+//
+// 		Result:
+// 		   URL - /v1/users/{userId}/details
+// 		   Composed URL - /v1/users/sample@sample.com/details
+// It replaces the value of the key while composing the request URL.
+//
+// Also it can be overridden at request level Path Params options,
+// see `Request.SetPathParam` or `Request.SetPathParams`.
+func (c *Client) SetPathParam(param, value string) *Client {
+	c.pathParams[param] = value
+	return c
+}
+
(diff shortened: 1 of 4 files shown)


moorereason (approved):
LGTM 👍

Issue #408, opened 2021-02-01
Execute user defined middlewares after resty's internal middlewares produced backward incompatible changes


buglloc:
Latest v2.4.0 release have backward incompatible changes: #355 
In this regard, I want to clarify - that now we can't manipulate the final request through `*resty.Request` is expected behavior or not? And now we need to rewrite all of our `OnBeforeRequest` middlewares?

Reproducer:
```
package main

import (
	"fmt"
	"io"
	"net/http"
	"net/http/httptest"
	"os"

	"github.com/go-resty/resty/v2"
)

func main() {
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		_, _ = io.Copy(os.Stdout, r.Body)
		_, _ = fmt.Fprintln(w, "LGTM")
	}))
	defer srv.Close()

	_, _ = resty.New().
		OnBeforeRequest(func(_ *resty.Client, r *resty.Request) error {
			r.Body = []int{1, 2, 3}
			return nil
		}).R().
		SetBody([]int{0, 0, 0}).
		Post(srv.URL)
}
```

Before v2.4.0 (this behavior introduced in the #57 afaik) output:
```
[1,2,3]
```

v2.4.0:
```
[0,0,0]
```

moorereason:
I agree.  We should not have merged this breaking change into a minor release.

My recommendation would be to revert the change and issue a new minor release (v.2.5.0) ASAP.  Then we can open a new issue and discuss the problem #355 was trying to solve.

Cc: @jeevatkm

jeevatkm:
@buglloc I'm sorry for the delay. I agree with @moorereason, apologies for the inconvenience. I will try to revert and make a release tonight or tomorrow. We will decide later on what do to with revert change.

buglloc:
jeevatkm@ any news? :)
The longer we wait with reverting, the more people can starts using the new behavior :(

jeevatkm:
@buglloc @moorereason Revert has been done!

buglloc:
Thanks!

PR #409, opened 2021-02-02
move udBeforeRequest before beforeRequest again & preReqHooks instead…

yflau:
As #408 reported, many RequestMiddlewares will be invalid when move udBeforeRequest after beforeRequest introduced in v2.4.0, thus all RequestMiddlewares have to change to modify RawReqest directly while not the Request. In the mean time, I noticed that for RequestMiddleware in v2.3.0, there is no way to access or manipulate the RawRequest, but there is another hook: PreRequestHook can meet the requirement, but I think it's better to support a list of PeRequestHook, so I modify the preRequestHook to preRequestHooks.

### DIFF
--- a/client.go
+++ b/client.go
@@ -130,7 +130,7 @@ type Client struct {
 	proxyURL           *url.URL
 	beforeRequest      []RequestMiddleware
 	udBeforeRequest    []RequestMiddleware
-	preReqHook         PreRequestHook
+	preReqHooks        []PreRequestHook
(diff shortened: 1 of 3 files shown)


jeevatkm (changes requested):
@yflau Do you mind reverting the changes of this #414? keeping other changes

yflau:
I think it's OK for PR #414 since the main purpose is reverting the breaking changes.

PR #413, opened 2021-02-10
Added a way set a Non-Canonical Header

Lao-Tuo:
Some API require use of Non-Canonical Header fields.

### DIFF
--- a/request.go
+++ b/request.go
@@ -117,6 +117,19 @@ func (r *Request) SetHeaders(headers map[string]string) *Request {
 	return r
 }
 
+// SetHeaderVerbatim method is to set a single header field and its value verbatim in the current request.
+//
+// For Example: To set `all_lowercase` and `UPPERCASE` as `available`.
+// 		client.R().
+//			SetHeader("all_lowercase", "available").
+//			SetHeader("UPPERCASE", "available")
+//
+// Also you can override header value, which was set at client instance level.
+func (r *Request) SetHeaderVerbatim(header, value string) *Request {
+	r.Header[header] = []string{value}
+	return r
+}
+
 // SetQueryParam method sets single parameter and its value in the current request.
 // It will be formed as query string for the request.
 //
--- a/request.go
+++ b/request.go
@@ -125,7 +125,7 @@ func (r *Request) SetHeaders(headers map[string]string) *Request {
 //			SetHeaderVerbatim("UPPERCASE", "available")
 //
 // Also you can override header value, which was set at client instance level.
-func (r *Request) SetHeaderVerbatim(header string, value string) *Request {
+func (r *Request) SetHeaderVerbatim(header, value string) *Request {
 	r.Header[header] = []string{value}
 	return r
 }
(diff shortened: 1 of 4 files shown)


moorereason:
I don't like this API.  Just set the map values manually and move on:

```
r := client.R()
r.Header["My-WHATEVER"] = "foo"
```

Lao-Tuo:
> I don't like this API. Just set the map values manually and move on:
> 
> ```
> r := client.R()
> r.Header["My-WHATEVER"] = "foo"
> ```

That should be:
r := client.R()
r.Header["My-WHATEVER"]  = []string{"foo"}

but sure, you can do that..

However, that would be the same with setting normal headers, too.. you can just skip the wrapping.
r := client.R()
r.Header.Set("My-WHATEVER","foo")

why do we have a "SetHeader" in the first place?

moorereason:
> why do we have a "SetHeader" in the first place?

To make the common case chainable with the API.

However, I'm softening to the idea of expanding the API for this.  I may have been in a cantankerous mood yesterday.

Lao-Tuo:
> To make the common case chainable with the API.

And that was the intention behind this one as well. ;)
The actual usecase that came up for me is the Cisco DNAC API (which needs non-canonical, lowercase Header-fields):
`
	response, err := RestyClient.R().
		SetHeaderVerbatim("entity_type", getClientEnrichmentDetailsQueryParams.Type).
		SetHeaderVerbatim("entity_value", getClientEnrichmentDetailsQueryParams.Value).
		SetResult(&GetClientEnrichmentDetailsResponse{}).
		SetError(&Error{}).
		Get(path)
`

request.go
```diff
@@ -117,6 +117,19 @@ func (r *Request) SetHeaders(headers map[string]string) *Request {
 	return r
 }
 
+// SetHeaderVerbatim method is to set a single header field and its value verbatim in the current request.
+//
+// For Example: To set `all_lowercase` and `UPPERCASE` as `available`.
+// 		client.R().
+//			SetHeader("all_lowercase", "available").
```
jeevatkm, line 124: @Lao-Tuo Please update godoc to the correct method

request.go
```diff
@@ -117,6 +117,19 @@ func (r *Request) SetHeaders(headers map[string]string) *Request {
...
+//			SetHeader("UPPERCASE", "available")
+//
+// Also you can override header value, which was set at client instance level.
+func (r *Request) SetHeaderVerbatim(header, value string) *Request {
```
jeevatkm, line 128: @Lao-Tuo Please add the same nature of the method into the Resty client struct too. Since resty provides most methods at both client and request level.

jeevatkm (changes requested):
@Lao-Tuo Thanks for your contributions. Can you please take care of the following items?
* Taking care of review comments
* Adding test case for new methods

I'm really sorry for the delay in PR review, due to personal reasons.

jeevatkm (approved):

PR #414, opened 2021-02-11
Reverting the breaking change introduced in the minor version #408

jeevatkm:


jeevatkm:
@moorereason I'm already late in doing this one. Once Travis's build is a success, will merge it.

PR #417, opened 2021-02-21
Feat: Add Retry hooks

### LINKED ISSUE #416: Proposal: OnRetry hook
Currently the Retry interface of the Resty client supports a custom condition for when to retry and a custom callback to calculate the time between retries. What I am currently missing is an `OnRetry` hook to execute side effecting operations whenever a request is retried.

For example, I have a use case where I want to authenticate with an API after an access token has expired. One potential way to solve that with `resty` would be to authenticate with the API if the request returned a 403. While it is currently possible to achieve this with the `resty.Backoff`, it would be much cleaner to have an `OnRetry` hook.

Blokje5:
Add RetryHooks: side-effecting functions that will be executed whenever a retry is triggered.

RetryHooks give users the ability to add side-effecting operations whenever a retry is triggered, without having to abuse the RetryConditions or using the Backoff directly. For example, RetryHooks allow you to log whenever a retry is triggered.

Fixes #416

### DIFF
--- a/client.go
+++ b/client.go
@@ -108,6 +108,7 @@ type Client struct {
 	RetryWaitTime         time.Duration
 	RetryMaxWaitTime      time.Duration
 	RetryConditions       []RetryConditionFunc
+	RetryHooks            []OnRetryFunc
 	RetryAfter            RetryAfterFunc
 	JSONMarshal           func(v interface{}) ([]byte, error)
 	JSONUnmarshal         func(data []byte, v interface{}) error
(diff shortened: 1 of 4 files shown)


jeevatkm (changes requested):
@Blokje5 Thanks for your contributions. PR looks good. Can you please resolve the PR conflicts?

Blokje5:
Thanks for taking the time to review the PR. I resolved the conflicts.

jeevatkm (approved):

PR #421, opened 2021-03-16
Don't use sync.Pool for json.NewEncoder target buffers

pborzenkov:
The slice returned by `noescapeJSONMarshal` continues to be accessed
even after `pool.Put` was called on the buffer. This might lead to
incorrect data being sent in request body if the buffer is acquired and
re-written by a concurrently running goroutine.

### DIFF
--- a/request.go
+++ b/request.go
@@ -838,11 +838,14 @@ func (r *Request) initValuesMap() {
 	}
 }
 
-var noescapeJSONMarshal = func(v interface{}) ([]byte, error) {
+var noescapeJSONMarshal = func(v interface{}) (*bytes.Buffer, error) {
 	buf := acquireBuffer()
-	defer releaseBuffer(buf)
 	encoder := json.NewEncoder(buf)
 	encoder.SetEscapeHTML(false)
-	err := encoder.Encode(v)
-	return buf.Bytes(), err
+	if err := encoder.Encode(v); err != nil {
+		releaseBuffer(buf)
+		return nil, err
+	}
+
+	return buf, nil
 }
--- a/client.go
+++ b/client.go
@@ -833,7 +833,6 @@ func (c *Client) GetClient() *http.Client {
 // Executes method executes the given `Request` object and returns response
 // error.
 func (c *Client) execute(req *Request) (*Response, error) {
-	defer releaseBuffer(req.bodyBuf)
 	// Apply Request middleware
 	var err error
 
@@ -866,6 +865,8 @@ func (c *Client) execute(req *Request) (*Response, error) {
 	if err = requestLogger(c, req); err != nil {
 		return nil, wrapNoRetryErr(err)
 	}
+
+	req.RawRequest.Body = newRequestBodyReleaser(req.RawRequest.Body, req.bodyBuf)
 
 	req.Time = time.Now()
 	resp, err := c.httpClient.Do(req.RawRequest)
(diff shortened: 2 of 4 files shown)


request.go
```diff
@@ -839,9 +839,8 @@ func (r *Request) initValuesMap() {
 }
 
 var noescapeJSONMarshal = func(v interface{}) ([]byte, error) {
-	buf := acquireBuffer()
```
jeevatkm, line 842: @pborzenkov do you mean usage of `sync.Pool` with `json.NewEncoder` has an issue at language level?

pborzenkov: @jeevatkm No. What I mean is that `bytes.Buffer` is returned back to `sync.Pool` before `noescapeJSONMarshal` returns (due to `defer`). Yet the underlying byte slice is used later, as `bytes.Buffer.Bytes()` doesn't copy the data.

So a concurrently running goroutine can get the very same `bytes.Buffer` from the pool and overwrite its contents before the data produced by `json.NewEncoder` is actually used.

jeevatkm: @pborzenkov Thanks. My understanding is `defer` executes after the `return` statement executed. Maybe I have to read,  try it out, and then get back to you.
https://golang.org/ref/spec#Defer_statements

pborzenkov: @jeevatkm 

Yes, it will execute after the `return`. The problem is that `return buf.Bytes()` does not produce a copy of the data, it return the underlying slice. And since the `buf` is returned back to the pool it can easily get reused.

jeevatkm: @pborzenkov I think I missed this part "doesn't copy the data" from your previous response. I want to keep the `sync.Pool` usage instead of creating new. It's better to restructure the method flow to make use of the pool properly.
Thanks for bringing it up 👍

pborzenkov: @jeevatkm 

Yeah, that's probably a good idea. Thanks!

BTW, there is another problem with how `sync.Pool` is used. When a request is made by the `execute` method, there is `defer` that releases request body back to the pool:

```
// Executes method executes the given `Request` object and returns response
// error.
func (c *Client) execute(req *Request) (*Response, error) {
	defer releaseBuffer(req.bodyBuf)
	// Apply Request middleware
```

But since on the first entry to this function (e.g. not a request retry) the `bodyBuf` is `nil`, the call to `releaseBuffer` is basically a no-op. The trivial fix would be to do it like this:

```
// Executes method executes the given `Request` object and returns response
// error.
func (c *Client) execute(req *Request) (*Response, error) {
	defer func() {
		releaseBuffer(req.bodyBuf)
	}()
	// Apply Request middleware
```

But this is probably racy, as `RoundTripper` may hold the body even after `http.Client.Do` returns:

>         // RoundTrip should not modify the request, except for
>         // consuming and closing the Request's Body. RoundTrip may
>         // read fields of the request in a separate goroutine. Callers
>         // should not mutate or reuse the request until the Response's
>         // Body has been closed.
>         //
>         // RoundTrip must always close the body, including on errors,
>         // but depending on the implementation may do so in a separate
>         // goroutine even after RoundTrip returns. This means that
>         // callers wanting to reuse the body for subsequent requests
>         // must arrange to wait for the Close call before doing so.

So the correct fix is to wrap `Response.Body` and release the request buffer whenever the response body is closed.

jeevatkm: @pborzenkov Your explanation and details make sense. Do you want to take a shot at it in this PR? or I will try it when I get a chance in the upcoming days.

pborzenkov: @jeevatkm Yeah. Gonna try and fix it.

jeevatkm: @pborzenkov I'm sorry for the delayed response at my end. Just went through your updates, it looks good. Also thanks for doing other improvements along with sync pool enhancement.

PR #428, opened 2021-05-11
Update Bazel build to work with 4.0.0

tanyabouman:
This includes several updates to the BUILD and WORKSPACE files, so that they now work with Bazel 4.0.0, Gazelle 0.23 and rules_go 0.27.

By default, [Gazelle now uses](https://github.com/bazelbuild/bazel-gazelle/pull/863) the last component of the Bazel package, rather than `go_default_library` as the `go_library name`.

This causes problems when the naming conventions do not match.
https://github.com/bazelbuild/bazel-gazelle/issues/890#issuecomment-685145667

Adding support for the new naming convention should not break anything that depends on using the old naming convention, because the Gazelle option `import_alias` makes aliases to the libraries and this allows the old naming convention to still work.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,36 +1,48 @@
-package(default_visibility = ["//visibility:private"])
-
-load("@bazel_gazelle//:def.bzl", "gazelle")
 load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+load("@bazel_gazelle//:def.bzl", "gazelle")
 
-gazelle(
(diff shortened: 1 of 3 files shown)


jeevatkm:
@tanyabouman Travis CI build link was broken. I have migrated travis-ci.org to travis-ci.com

Can you please edit the PR or close & reopen it to trigger the CI build?

jeevatkm (approved):
@tanyabouman Thank you for your PR and contribution.

PR #436, opened 2021-07-14
Added RetryConditions to Request

### LINKED ISSUE #315: is there any reason for retries not to be configured per request?
Hi there,
I noticed that retries are configured per client, 
is there any reason not to be per request?

Should I always create a new client each time I want to change RetryCount, RetryMaxWaitTime or RetryCondition without affecting one another if it runs concurrently?

For example:
```
defaultClient := resty.New()


func One() error{
client.
   defaultClient.SetRetryCount(3).SetRetryMaxWaitTime(20 * time.Second)
   defaultClient.R().Post........
}

func Two() error{
client.
   defaultClient.SetRetryCount(5).SetRetryMaxWaitTime(10* time.Second)
   defaultClient.R().Post........
}
```
in that case I should create a new instance for each function, right ?

### LINKED ISSUE #324: Request level retry strategy support
Is there plan to support request level retry strategy? Currently request will use client retry strategy, and can not override at request level.

### LINKED ISSUE #433: Request Specific Retry Conditions
Currently Retry Conditions are added to the client which are then subsequently used by the request. This introduces the need to create a new client to create a request with different retry condition. A workaround would be setting`client.RetryConditions` to an empty slice, which would fail is used concurrently.

I propose Request Specific RetryConditions, while keeping the client one and using combination of both. Will make a draft PR soon!

rohitkg98:
- Retry conditions can now also be added to Request

Closes #433 
Closes #315
Closes #324

### DIFF
--- a/request.go
+++ b/request.go
@@ -67,6 +67,7 @@ type Request struct {
 	clientTrace         *clientTrace
 	multipartFiles      []*File
 	multipartFields     []*MultipartField
+	retryConditions     []RetryConditionFunc
 }
 
 // Context method returns the Context if its already set in request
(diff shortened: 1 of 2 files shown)


rohitkg98:
@jeevatkm any updates on this?

request.go
```diff
@@ -747,7 +758,7 @@ func (r *Request) Execute(method, url string) (*Response, error) {
 		Retries(r.client.RetryCount),
 		WaitTime(r.client.RetryWaitTime),
 		MaxWaitTime(r.client.RetryMaxWaitTime),
-		RetryConditions(r.client.RetryConditions),
+		RetryConditions(append(r.client.RetryConditions, r.retryConditions...)),
```
jeevatkm, line 761: @rohitkg98 If we are adding a Retry condition feature at the request level; is it a request retry condition that should take place first instead of the client?

@moorereason your thoughts!

jeevatkm (changes requested):

request.go
jeevatkm: @rohitkg98 Basically, Request retry conditions have to execute first and then client instance level. So we need to flip the append arguments.

chb0github: I am looking for this feature as well

rohitkg98: changed to check request specific conditions before client ones

chb0github:
Stoked for a merge

chb0github:
??

jeevatkm (approved):
@rohitkg98 Thanks for the PR and contribution.

chb0github:
When will this be released? I updated my go.mod dependency to 2.7.0 and it isn't resolving

PR #451, opened 2021-08-27
fix add pre request hook return new request

hhy5861:
fix open tracing pre request hook return new request

### DIFF
--- a/client.go
+++ b/client.go
@@ -74,7 +74,7 @@ type (
 	ResponseMiddleware func(*Client, *Response) error
 
 	// PreRequestHook type is for the request hook, called right before the request is sent
-	PreRequestHook func(*Client, *http.Request) error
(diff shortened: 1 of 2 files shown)


hhy5861:
> I would recommend opening an issue to discussion this breaking change to the API.

`func (opt *Options) PreRequestHook(c *resty.Client, req *http.Request) (*http.Request, error) {
	spanName := fmt.Sprintf("%s %s: %s", operationName, req.Method, req.URL)

	httpRequest := req.WithContext(req.Context())
	httpRequest, _ = nethttp.TraceRequest(
		opt.tracer, httpRequest,
		nethttp.ClientTrace(true),
		nethttp.InjectSpanContext(true),
		nethttp.OperationName(spanName),
		nethttp.ComponentName(defaultComponentName),
	)

	return httpRequest, nil
}`

PR #476, opened 2021-10-14
Make PathParams public

lavoiesl:
This PR makes the `PathParams` property public, same as `QueryParam`, `FormData`, `Header`, etc.

I don't really see the value in having it private and it prevents me from adding a middleware to capture all the existing parameters.

Example:
```go
func InjectContextMiddleware(service string) resty.RequestMiddleware {
	return func(c *resty.Client, req *resty.Request) error {
		ctx := req.Context()
		ctx = logger.WithFields(ctx, map[string]interface{}{
			"http_service":         service,
			"http_service_host":    c.HostURL,
			"http_request_method":  req.Method,
			"http_request_uri":     req.URL, // URLs should have fairly low cardinality. Users should use PathParams and QueryParams for variable parts
			"http_request_attempt": req.Attempt,
			"http_request_retry":   req.Attempt > 1,
		})

		for k, v := range req.QueryParam {
			if len(v) > 1 {
				ctx = logger.WithField(ctx, "http_request_query_"+k, v)
			} else {
				// most common scenario, only one value; inline the value instead of serializing an array
				ctx = logger.WithField(ctx, "http_request_query_"+k, v[0])
			}
		}

		// Currently not possible
		for k, v := range req.PathParams {
			ctx = logger.WithField(ctx, "http_request_path_"+k, v)
		}

		req.SetContext(ctx)
		return nil
	}
}
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -95,6 +95,7 @@ type Client struct {
 	HostURL               string
 	QueryParam            url.Values
 	FormData              url.Values
+	PathParams            map[string]string
 	Header                http.Header
 	UserInfo              *User
 	Token                 string
@@ -125,7 +126,6 @@ type Client struct {
 	debugBodySizeLimit int64
 	outputDirectory    string
 	scheme             string
-	pathParams         map[string]string
 	log                Logger
 	httpClient         *http.Client
 	proxyURL           *url.URL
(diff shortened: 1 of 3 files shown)


jeevatkm (approved):
@lavoiesl Thanks for the test execution info.

PR #477, opened 2021-10-14
Add WrapExecutor

lavoiesl:
Add the ability to wrap the request executor, granting full access to the request, response, and error.
Runs on every request attempt, before any request hook and after any response or error hook.

Can be useful to introduce throttling or add hooks that always fire, regardless of success or error.
Having access to the request even in case of errors means also having access to the request's context, which can be immensely helpful for proper error logging/reporting.

This PR is dependant on https://github.com/go-resty/resty/pull/478, to avoid re-wrapping `noRetryError`

### DIFF
--- a/request.go
+++ b/request.go
@@ -755,8 +755,9 @@ func (r *Request) Execute(method, url string) (*Response, error) {
 
 	if r.client.RetryCount == 0 {
 		r.Attempt = 1
+		resp, err = r.client.executor(r)
 		r.client.onErrorHooks(r, resp, err)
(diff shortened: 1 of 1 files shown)


lavoiesl:
@jeevatkm Is there still a plan to release a v3? We ended up having to maintain a fork of resty because this hook proved essential to throttling, error handling, and metrics gathering.

jeevatkm:
@lavoiesl I will aim this concept for v3. I know, I got delayed so much, I will make up for it. Thanks for your patience.

PR #478, opened 2021-10-18
Change noRetryError for temporary errors

lavoiesl:
- Expose an `Unwrap` method to leverage go's standard unwrapping
- Let the error escape, such that hooks and end-users can access the `Temporary` information.
- Do not wrap if the error already exposes `Temporary`, to honour it.

The main benefit is to leverage `net/*`'s standard errors; most of them already expose `Temporary` methods.


jeevatkm:
@lavoiesl please do not close, I will be merging this PR for v3 tonight.

jeevatkm:
I will do resolve conflicts and merge.

jeevatkm:
Thanks, I will take care of it.

PR #502, opened 2021-12-21
Update README, add netrc BasicAuth

cnmade:


cnmade:
This PR is too old. i think nobody need it .
I am currently use an fork go-resty by myself, it reach my need. so i close this pr

PR #528, opened 2022-03-14
Fix Send() documentation code

lordzsolt:
I'm guessing the previous code was a typo and `client.R().Send()` would just send an empty request, not the configured `req`.


jeevatkm:
Thanks @lordzsolt for improving resty docs. I'm sorry for the delayed attention on the PR.

PR #541, opened 2022-05-06
Fix SetAllowGetMethodPayload option ignored for io.Reader payload

picollomartin:
Hi 👋 !

We detect some inconsistent behaviour of the current implementation of the  `SetAllowGetMethodPayload` based on the input type passed to `SetBody` option:

- SetAllowGetMethodPayload = false
  - Body type = io.Reader => body is sent in GET
  - Body type = non io.Reader => body is not sent in GET
   
 - SetAllowGetMethodPayload = true
   - Body type = io.Reader => body is sent in GET
   - Body type = non io.Reader => body is sent in GET
   
Basically you can bypass the option AllowGetMethodPayload if you send an io.Reader as a Body, this fixes that for be consistent to all types and add a test case for test disabled option

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -468,6 +468,21 @@ func TestClientAllowsGetMethodPayload(t *testing.T) {
 	assertEqual(t, payload, resp.String())
 }
 
+func TestClientAllowsGetMethodPayloadDisabled(t *testing.T) {
+	ts := createGetServer(t)
+	defer ts.Close()
+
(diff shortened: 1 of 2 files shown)


jeevatkm:
@picollomartin I'm sorry for the delayed attention on the PR. Thank you for your contribution. 
Agreed, inconsistent behavior will lead to consuming investigation time. This PR provided insight for #618.

Issue #166, opened 2018-06-26
Collecting Inputs for Resty v2.0 and Plan the release


jeevatkm:
The goal of this thread to collect feedback's from `resty` users and `plan v2.0.0` release. Action items get recorded here #167.

- Things could be improved
- Things could be added (generalized one's not specific use case)
- Things could be removed

Share your resty experiences for upcoming `v2.0.0`

moorereason:
I've used Resty for two small CLI projects that interact with 3rd party REST services.  Resty has done everything I wanted it to do and saved me a lot of time.  The Resty 1.6 API is adequate for me, but my needs are rather typical so far.

The only hurdles I had to overcome were trying to understand how to use Resty's features. Showing a typical usage example with error handling would have helped me better understand the Resty design at a glance.  For example, here's how I typically use Resty when interacting with the [Quizlet API](https://quizlet.com/api/2.0/docs):

```go
// GetSets returns all user sets in Quizlet.
func (c *Config) GetSets() ([]Set, error) {
	var sets []Set
	var e QuizletError

	uri := fmt.Sprintf("%s/users/%s/sets", apiBaseURL, c.Username)

	_, err := resty.SetDebug(c.Debug).R().
		SetHeader("Accept", "application/json").
		SetAuthToken(c.AuthToken).
		SetResult(&sets).
		SetError(&e).
		Get(uri)
	if err != nil {
		return nil, err
	}

	if e.Code != 0 {
		return nil, e
	}

	return sets, err
}

// QuizletError represents an error response from the Quizlet API.
type QuizletError struct {
	Code             int      `json:"http_code"`
	QError           string   `json:"error"`
	Title            string   `json:"error_title"`
	Description      string   `json:"error_description"`
	ValidationErrors []string `json:"validation_errors"`
}

// Error implements the error interface.
func (e QuizletError) Error() string {
	return e.Description
}
```

jeevatkm:
Thank you for your inputs @moorereason.

Action Item for v2.0 release:

* Improve documentation and examples

h7kanna:
How about having DefaultClient a DefaultTransport with sane defaults for Timeouts?
For example https://github.com/hashicorp/go-cleanhttp/blob/master/cleanhttp.go#L24

jeevatkm:
@h7kanna Thank you for your input, I have noted down here #167.

zjjhzx:
Any plans on supporting "net/http2" ?

jeevatkm:
[...]

PR #549, opened 2022-06-15
Fix multipart file readers not being reset when doing a retry

nikplx:
As mentioned in https://github.com/go-resty/resty/issues/166 by @neganovalexey, when doing a retry on a multipart request the `io.Reader` is not reset to the start again. This can lead to subtle bugs, as the request will be retried, but the form part of the file will be empty, because the reader was already exhausted. Solving this issue in the caller code is pretty ugly, because `Request` does not expose access to `multipartFiles` and so the caller would have to use `RetryHooks` and recreate the `Response.Request` and that at each call site (where there is access to the original `ReadSeeker`).

In the PR, I created a new client options `SetRetryResetReaders` which adds a retry hook  seeking the start of each reader in `multipartFiles`, if the reader it got also implements `io.ReadSeeker`. This solved the above issue described above for my usecase and should have minimal implications on existing usages.

### DIFF
--- a/client.go
+++ b/client.go
@@ -116,6 +116,7 @@ type Client struct {
 	RetryConditions       []RetryConditionFunc
 	RetryHooks            []OnRetryFunc
 	RetryAfter            RetryAfterFunc
+	RetryResetReaders     bool
 	JSONMarshal           func(v interface{}) ([]byte, error)
 	JSONUnmarshal         func(data []byte, v interface{}) error
 	XMLMarshal            func(v interface{}) ([]byte, error)
@@ -699,6 +700,15 @@ func (c *Client) AddRetryAfterErrorCondition() *Client {
 // Since v2.6.0
 func (c *Client) AddRetryHook(hook OnRetryFunc) *Client {
 	c.RetryHooks = append(c.RetryHooks, hook)
+	return c
+}
+
+// SetRetryResetReaders method enables the Resty client to seek the start of all
+// file readers given as multipart files, if the given object implements `io.ReadSeeker`.
+//
+// Since ...
+func (c *Client) SetRetryResetReaders(b bool) *Client {
+	c.RetryResetReaders = b
 	return c
 }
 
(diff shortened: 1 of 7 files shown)


SVilgelm:
What about the readers that cannot be reset?

nikplx:
@SVilgelm 
They won't be reset. The code only resets the reader if it implements `io.ReadSeeker` and that only if you enable it, so as to not break any running code. Every other reader will be untouched. I don't know if there is a usecase for passing in a reader that should not be reset, but it will not do anything if there is a reader that can not be reset.

jeevatkm:
@nikplx I'm sorry for the delayed attention on the PR. Thanks for your contribution.

client.go
```diff
@@ -638,6 +639,15 @@ func (c *Client) AddRetryHook(hook OnRetryFunc) *Client {
 	return c
 }
 
+// SetRetryResetReaders method enables the Resty client to seek the start of all
+// file readers given as multipart files, if the given object implements `io.ReadSeeker`.
+//
+// Since ...
```
jeevatkm, line 645: I will fill in the version in before the release.

jeevatkm:
@nikplx I'm sorry for the delayed attention on the PR. Thanks for your contribution.
Yes, it make sense to provide the reset readers option for resty users.

jeevatkm (changes requested):
@nikplx Can you please help to get the PR build validation cleared? then we can merge this PR.

nikplx:
Hey @jeevatkm 

I fixed the test stage. There was some mixup in the sum file. It now runs locally. Maybe you can retrigger the pipeline if you have some time.

jeevatkm:
@nikplx This time PR validation build executed without any issue.

Do you mind checking this [codecov feedback](https://github.com/go-resty/resty/pull/549/checks?check_run_id=11786008284)?

jeevatkm (approved):
Thanks @nikplx 👍

PR #555, opened 2022-06-28
Feat: support set querystring from struct

vlean:
`SetQueryStruct` method set struct parameter and its value in the current request.

For Example:
```go
 type Options struct {
  	Query   string `url:"q"`
  	ShowAll bool   `url:"all"`
  	Page    int    `url:"page,omitempty"`
 }
 opt := Options{Query:"foo", ShowAll: true}
 client.R().
	SetQueryStruct(opt)
```

will output: "q=foo&all=true"

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,8 @@
 module github.com/go-resty/resty/v2
 
-require golang.org/x/net v0.0.0-20211029224645-99673261e6eb
+require (
+    github.com/google/go-querystring v1.1.0
+    golang.org/x/net v0.0.0-20211029224645-99673261e6eb
+)
 
 go 1.11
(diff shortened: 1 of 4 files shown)


SVilgelm:
will be great to have same feature for path params as well

moorereason:
I would rather have the user implement this simple helper function themselves instead of adding an external dependency to resty.

vlean:
> I would rather have the user implement this simple helper function themselves instead of adding an external dependency to resty.

hi, Is it possible if I copy the `query.Values` code to the repository?
it will solve the chained call terminal. 

Before: 
You have to add this code every where you use the GET method
```go
client.R().SetQueryParamsFromValues(helper.Query(OptStruct))
```

After:
```go
client.R().SetQueryStruct(OptStruct)
```

moorereason:
> Is it possible if I copy the `query.Values` code to the repository?

I assume you mean adding this to a "helper" sub-package within resty.  I don't personally think it's worth the trouble, but perhaps others would disagree.

Also, I'd prefer to have this discussion in a feature request issue instead of a PR.

PR #568, opened 2022-07-26
Fix redirect request body mismatch with origin request body caused by buffer reused by mistake

### LINKED ISSUE #567: Bug report: Request.bodyBuf resused by mistake
# Summary
Resty [saved the request body's snapshot](https://github.com/go-resty/resty/blob/master/middleware.go#L521) for resending request with original HTTP method and body when http redirect such as 307 or 308 returned.
Then in [http.Client.do](https://github.com/golang/go/blob/master/src/net/http/client.go#L661), golang call the GetBody func registered by resty [createHTTPRequset](https://github.com/go-resty/resty/blob/master/middleware.go#L215) func for reset the request body. 

But the [code](https://github.com/go-resty/resty/blob/master/middleware.go#L521) is not the body copy.
```
func getBodyCopy(r *Request) (*bytes.Buffer, error) {
	// If r.bodyBuf present, return the copy
	if r.bodyBuf != nil {
		return bytes.NewBuffer(r.bodyBuf.Bytes()), nil
	}
```
bodyCopy and Request.bodyBuf shared the same bytes slice in go. Once HTTP redirect happened, request body(requestBodyReleaser) was closed and released by go in [http client](https://github.com/golang/go/blob/master/src/net/http/client.go#L733). The buffer return to bufPool and may be reused by other requests. If other requests rewrite this buffer, the origin bodyCopy is overwrited also and then the redirected request body passed wrong body mismatch with the original HTTP request.

# Reproduce

```
package main

import (
	"fmt"
	"io/ioutil"
	"math/rand"
	"net/http"
	"sync"
	"time"

)

var data = []byte("0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")

func main() {
	client := resty.New()
	server := http.NewServeMux()
	server.HandleFunc("/test", func(writer http.ResponseWriter, request *http.Request) {
		writer.Header().Set("Location", "http://localhost:9999/redirect")
		writer.WriteHeader(http.StatusTemporaryRedirect)
		writer.Write(nil)
	})
	server.HandleFunc("/redirect", func(writer http.ResponseWriter, request *http.Request) {
		body, err := ioutil.ReadAll(request.Body)
		if err != nil {
			panic(err)
		}
		key := request.Header.Get("key")
		if key != string(body) {
			fmt.Println("mismatch key and body", key, string(body))
		}
		writer.WriteHeader(http.StatusOK)
	})

	go func() {
		err := http.ListenAndServe("localhost:9999", server)
		if err != nil {
			panic(err)
		}
	}()

	wg := sync.WaitGroup{}
	for i := 0; i < 2; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			post(client)
		}()
	}
	wg.Wait()
}

func post(client *resty.Client) {
	for {
		key := RandStr(64)
		client.NewRequest().SetHeader("key", key).SetBody([]byte(key)).Post("http://localhost:9999/test")
		time.Sleep(10 * time.Millisecond)
	}
}

func RandStr(length int) string {
	result := []byte{}
	randCli := rand.New(rand.NewSource(time.Now().UnixNano()))
	for i := 0; i < length; i++ {
		result = append(result, data[randCli.Intn(len(data))])
	}
	return string(result)
}
```
Expected key's value in header is always same as body, but appear mismatch by run the program above.

image

liguangbo:
As issue #567 described.

Copy the body buffer by buffer.Write to fix body mismatch with origin request when 307 or 308 redirect.

closes #567

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -17,6 +17,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 )
@@ -857,4 +858,28 @@ func TestHostURLForGH318AndGH407(t *testing.T) {
 		Post("/login")
 	assertNil(t, err)
 	assertNotNil(t, resp)
+}
+
+func TestPostRedirectWithBody(t *testing.T) {
+	ts := createPostServer(t)
+	defer ts.Close()
+
+	targetURL, _ := url.Parse(ts.URL)
+	t.Log("ts.URL:", ts.URL)
+	t.Log("targetURL.Host:", targetURL.Host)
+
+	c := dc()
+	wg := sync.WaitGroup{}
+	for i := 0; i < 100; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			resp, err := c.R().
+				SetBody([]byte(strconv.Itoa(newRnd().Int()))).
+				Post(targetURL.String() + "/redirect-with-body")
+			assertError(t, err)
+			assertNotNil(t, resp)
+		}()
+	}
+	wg.Wait()
 }
--- a/middleware.go
+++ b/middleware.go
@@ -514,7 +514,7 @@ func saveResponseIntoFile(c *Client, res *Response) error {
 func getBodyCopy(r *Request) (*bytes.Buffer, error) {
 	// If r.bodyBuf present, return the copy
 	if r.bodyBuf != nil {
-		bodyCopy := &bytes.Buffer{}
+		bodyCopy := acquireBuffer()
 		if _, err := io.Copy(bodyCopy, bytes.NewReader(r.bodyBuf.Bytes())); err != nil {
 			// cannot use io.Copy(bodyCopy, r.bodyBuf) because io.Copy reset r.bodyBuf
 			return nil, err
--- a/middleware.go
+++ b/middleware.go
@@ -514,7 +514,12 @@ func saveResponseIntoFile(c *Client, res *Response) error {
 func getBodyCopy(r *Request) (*bytes.Buffer, error) {
 	// If r.bodyBuf present, return the copy
 	if r.bodyBuf != nil {
-		return bytes.NewBuffer(r.bodyBuf.Bytes()), nil
+		bodyCopy := &bytes.Buffer{}
+		if _, err := io.Copy(bodyCopy, bytes.NewReader(r.bodyBuf.Bytes())); err != nil {
+			// cannot use io.Copy(bodyCopy, r.bodyBuf) because io.Copy reset r.bodyBuf
+			return nil, err
+		}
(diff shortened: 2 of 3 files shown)


liguangbo:
Anyone can help？ 
It's a serious problem in some cases, such as user A POST user B's data to server, and lost user A's data.

@jeevatkm

thalesfsp:
@jeevatkm Sept 22

client_test.go
```diff
@@ -787,3 +788,27 @@ func TestHostURLForGH318AndGH407(t *testing.T) {
...
+	c := dc()
+	wg := sync.WaitGroup{}
+	for i := 0; i < 100; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			resp, err := c.R().
+				SetBody([]byte(strconv.Itoa(newRnd().Int()))).
+				Post(targetURL.String() + "/redirect-with-body")
+			assertNil(t, err)
```
segevda, line 809: you can use assertError instead of assertNil here

resty_test.go
```diff
@@ -291,6 +292,19 @@ func createPostServer(t *testing.T) *httptest.Server {
 				w.Header().Set(hdrLocationKey, "/login")
 				w.WriteHeader(http.StatusTemporaryRedirect)
 			}
+
```
segevda, line 295: maybe switch on r.URL.Path (like what's done on createGetServer)

segevda (changes requested):
Very nice catch!

middleware.go
```diff
@@ -521,7 +521,9 @@ func saveResponseIntoFile(c *Client, res *Response) error {
 func getBodyCopy(r *Request) (*bytes.Buffer, error) {
 	// If r.bodyBuf present, return the copy
 	if r.bodyBuf != nil {
-		return bytes.NewBuffer(r.bodyBuf.Bytes()), nil
+		bodyCopy := &bytes.Buffer{}
+		bodyCopy.Write(r.bodyBuf.Bytes())
```
segevda, line 525: there's an unhandled error here. Anyway, I think it's clearer if we use io.Copy here instead

liguangbo:
@segevda thanks for the review

segevda (approved):

bbrodriges:
@segevda will this fix be merged? This bug affects our production code.

segevda:
@bbrodriges unfortunately I don't have merge capabilities for this repo. Maybe @jeevatkm or @moorereason can take a look..?

jeevatkm (approved):
Thanks @liguangbo, I have updated the line.

liguangbo:
@jeevatkm Sorry for didn't see the notification. Happy to contribute to resty.

PR #582, opened 2022-09-21
feat: add nil check to SetResult (#503)

### LINKED ISSUE #503: SetResult Add nil Check
SetResult Add nil Check, avoid panic

tkrop:
fixes #503.

Simply adds a `nil` check to `SetResult` to ensure that it can be used fluently. Also adds a simple unit test to ensure that the current panic will not happen any longer.

### DIFF
--- a/request.go
+++ b/request.go
@@ -303,7 +303,9 @@ func (r *Request) SetBody(body interface{}) *Request {
 // Accessing a result value from response instance.
 //		response.Result().(*AuthToken)
 func (r *Request) SetResult(res interface{}) *Request {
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -303,7 +303,9 @@ func (r *Request) SetBody(body interface{}) *Request {
 // Accessing a result value from response instance.
```
segevda, line 303: I wonder if this should be documented here? something like
`passing a nil res paramaeter has no effect`

jeevatkm: @segevda I will add it later it. Thanks!

PR #583, opened 2022-09-28
Add digest authentication

### LINKED ISSUE #572: Does resty support digest authentication
Does resty support digest authentication or is there any plan for it

segevda:
- Adhere to RFC 7616 ("HTTP Digest Access Authentication")
- Added SetDigestAuth methods for Client and Request
- Currently not supporting auth-int Quality of Protection

closes #467
closes #572

### DIFF
--- a/digest.go
+++ b/digest.go
@@ -14,11 +14,11 @@ import (
 )
 
 var (
-	ErrBadChallenge    = errors.New("challenge is bad")
-	ErrCharset         = errors.New("unsupported charset")
-	ErrAlgNotSupported = errors.New("algorithm is not supported")
-	ErrQopNotSupported = errors.New("no supported qop in list")
-	ErrNoQop           = errors.New("qop must be specified")
+	ErrDigestBadChallenge    = errors.New("digest: challenge is bad")
+	ErrDigestCharset         = errors.New("digest: unsupported charset")
+	ErrDigestAlgNotSupported = errors.New("digest: algorithm is not supported")
+	ErrDigestQopNotSupported = errors.New("digest: no supported qop in list")
+	ErrDigestNoQop           = errors.New("digest: qop must be specified")
 )
 
 var hashFuncs = map[string]func() hash.Hash{
@@ -56,7 +56,7 @@ func (dt *digestTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 	}
 	chal := resp.Header.Get(hdrWwwAuthenticateKey)
 	if chal == "" {
-		return resp, ErrBadChallenge
+		return resp, ErrDigestBadChallenge
 	}
 
 	c, err := parseChallenge(chal)
@@ -113,7 +113,7 @@ func parseChallenge(input string) (*challenge, error) {
 	const qs = `"`
 	s := strings.Trim(input, ws)
 	if !strings.HasPrefix(s, "Digest ") {
-		return nil, ErrBadChallenge
+		return nil, ErrDigestBadChallenge
 	}
 	s = strings.Trim(s[7:], ws)
 	sl := strings.Split(s, ", ")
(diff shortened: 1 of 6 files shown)


digest.go
```diff
@@ -0,0 +1,295 @@
...
+		sl = append(sl, fmt.Sprintf("nc=%08x", c.nc))
+		sl = append(sl, fmt.Sprintf(`cnonce="%s"`, c.cNonce))
+	}
+
+	return fmt.Sprintf("Digest %s", strings.Join(sl, ", ")), nil
+}
+
+func (c *credentials) validateQop() error {
+	// Currently only supporting auth quality of protection. TODO: add auth-int support
+	// NOTE: cURL support auth-int qop for requests other than POST and PUT (i.e. w/o body) by hashing an empty string
```
jeevatkm, line 216: @segevda auth-int can be done later as a separate PR.

digest.go
```diff
@@ -0,0 +1,295 @@
+package resty
```
jeevatkm, line 1: @segevda Did you use any code reference or documentation for this implementation? If yes, please add those references in the code file.

jeevatkm (changes requested):
@segevda Thank you of your PR and contribution. I'm sorry for the delayed attention on the PR.
Can you please check the review comments?

digest.go
segevda: Yes I agree. I think this comment should remain for reference. The cURL implementation is something worth considering

digest.go
segevda: This code was loosely inspired by [bobziuchkovski/digest](https://github.com/bobziuchkovski/digest). I don't mind referencing it, just not sure if this has any license implications. The code in that repo is under Apache 2.0 and resty is under MIT. WDYT?

jeevatkm: I see, thanks for the info.

jeevatkm (approved):
@segevda Thanks for the PR.

segevda:
> @segevda It seems test cases coverage check is failing, can you please check it?

@jeevatkm just to make sure I understand, the failure here is because the total codebase coverage dropped due to this addition?

jeevatkm:
> > @segevda It seems test cases coverage check is failing, can you please check it?
> 
> @jeevatkm just to make sure I understand, the failure here is because the total codebase coverage dropped due to this addition?

Yes, @segevda.

segevda:
@jeevatkm I don't see how I can improve the coverage on this file any more. All missing lines are error checks for std lib functions. Please assist.

jeevatkm:
@segevda I see, I believe you have looked at this page. Which lists those return error lines, correct?
https://github.com/go-resty/resty/pull/583/checks?check_run_id=11833912592 

Possibly, we could supply incorrect input values to produce the error in the test! what are you thoughts?

segevda:
@jeevatkm I found a way to reduce error checking and thus to improve coverage:
Hash.Write calls are documented to never return an error ([see here](https://pkg.go.dev/hash#Hash)).
Also, changed the way we assign the digest Transport to avoid data races.

Please review :pray:

digest.go
```diff
@@ -0,0 +1,272 @@
...
+	"strings"
+)
+
+var (
+	ErrBadChallenge    = errors.New("challenge is bad")
+	ErrCharset         = errors.New("unsupported charset")
+	ErrAlgNotSupported = errors.New("algorithm is not supported")
+	ErrQopNotSupported = errors.New("no supported qop in list")
+	ErrNoQop           = errors.New("qop must be specified")
+)
```
jeevatkm, line 22: @segevda Can you please update Error names to following?
```go
var (
	ErrDigestBadChallenge    = errors.New("digest: challenge is bad")
	ErrDigestCharset         = errors.New("digest: unsupported charset")
	ErrDigestAlgNotSupported = errors.New("digest: algorithm is not supported")
	ErrDigestQopNotSupported = errors.New("digest: no supported qop in list")
	ErrDigestNoQop           = errors.New("digest: qop must be specified")
)
```

jeevatkm (changes requested):
@segevda Thanks for improving the PR in the last iteration. I have one last comment, can you have a look?

digest.go
segevda: Done

jeevatkm (approved):
Thanks @segevda for resolving the comments

PR #586, opened 2022-10-01
add OnSuccess, OnPanic, and OnInvalid hooks

muir:
Working to integrate resty with [xop](https://github.com/muir/xop-go), I need to be able to detect when a request is finished.  `OnError` already tracks one way a request could finish, but there are three other ways.

My need would be replicated by anyone else working to integrate resty with span/trace-based logging.


jeevatkm (approved):
@muir Thanks for your PR and contribution. I'm sorry for the delayed attention on the PR.

PR #590, opened 2022-10-19
feat: add SetResolver method support custom dns server

dcboy:


client.go
```diff
@@ -658,6 +660,31 @@ func (c *Client) SetTLSClientConfig(config *tls.Config) *Client {
 	return c
 }
 
+// SetResolver method set Custom DNS for resolver
```
segevda, line 663: The comment should be more informative IMO. please include an example and also address the fact that this setting will be overridden if SetTransport is called after, like so:
`c.SetResolver("foo").SetTransport(someOtherTransport)`

segevda (commented):
Can you please elaborate on why this is required? Sound like a specific edge case that can be solved by setting the transport via c.SetTransport()

jeevatkm:
@dcboy Thanks for your PR and contribution. I'm sorry for the delayed attention on the PR.

I have looked at the PR. Typically adding a custom DNS resolver is less common also it could be easily achieved with custom transport. I would suggest to create a separate PR for updating Resty readme with your example.

Issue #594, opened 2022-10-25
SetResult fails to parse JSON when the returned Content-Type is "application/x-amz-json-1.1" 


zackwine:
I found the SetResult API fails to parse the result JSON if the `Content-Type` is equal to "application/x-amz-json-1.1" (which Amazon uses for some endpoints).  Would it be possible to add this as a supported Content-Type, or possibly support setting a custom JSON Content-Type for resty?

Looks like this is the regex that is failing to match:  https://github.com/go-resty/resty/blob/master/client.go#L63

segevda:
As a workaround, you can set `application/json` as the content-type using Request.ForceContentType method, which will force resty to unmarshall the resp body as a JSON.

zackwine:
@segevda  Thanks!

PR #595, opened 2022-10-31
feat: Enable to customize unmarshal rule

tsbkw:
I found issue #594 and after looking into it I come up with the idea to deal with it.
I hope you to give me some feedback, because I'm not familiar with this repository.

[Issue #594](#594) can be worked around by set `application/json` as ForceContentType, but content-type may be different depends on response, and ForceContentType assumes that server always return same content type.

AddForceUnmarshalRule enables unmarshal depends on content-type from server.

Note1: Previously createGenServer return xml response without root element and it cannot be unmarshalled because encoding/xml.Unmarshal is only unmarshal chidlren of root element.
Now add root element in response and change assertion in test case accordingly.

Note2: Unmarshalc is no longer used in this repository but preserved for backward compatibility.

### DIFF
--- a/client.go
+++ b/client.go
@@ -390,6 +390,7 @@ func (c *Client) R() *Request {
 		multipartFields: []*MultipartField{},
 		PathParams:      map[string]string{},
 		jsonEscapeHTML:  true,
+		forceUnmarshal:  map[string]UnmarshalType{},
 	}
 	return r
 }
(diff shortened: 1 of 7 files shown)


jeevatkm:
@tsbkw Thank you for your PR and contribution. I'm sorry for the delayed attention on the PR.

I have looked at the PR. It seems, it is getting shallowed with force content type option also I have plans to add feature of registering a marshaler for content type in the Resty v3.

Regarding the #594 issue, at present I believe, [regex expression here](https://github.com/go-resty/resty/blob/master/client.go#L63) needs an attention.

So, I'm going to close this PR. I appreciate your attempt. Thanks!

PR #597, opened 2022-11-04
Fix race condition during request: Protect pre/post request hooks with RWMutex

jameshoulahan:
Firstly, thanks for the wonderful library! It's an integral part of one of my team's projects and it works beautifully.

In our project, we have one resty client that performs all API requests. It is shared among multiple "users" that each inject their own auth tokens into each request via pre request hooks.

The issue is, if one user is performing requests at the same time as another user is being created (which involves setting up its pre-request hooks), our CI reports a data race.

Given that an API client such as resty is intended to be used in a highly parallel way, potentially by many different places in the code, it makes sense to me that user-modifiable things such as pre- and post-request hooks should be protected from data races. As such, I have added an RWMutex for each user-modifiable hook slice.

### DIFF
--- a/client.go
+++ b/client.go
@@ -122,24 +122,26 @@ type Client struct {
 	// value when `SetAuthToken` option is used.
 	HeaderAuthorizationKey string
 
-	jsonEscapeHTML     bool
-	setContentLength   bool
-	closeConnection    bool
-	notParseResponse   bool
-	trace              bool
-	debugBodySizeLimit int64
-	outputDirectory    string
-	scheme             string
-	log                Logger
-	httpClient         *http.Client
-	proxyURL           *url.URL
-	beforeRequest      []RequestMiddleware
-	udBeforeRequest    []RequestMiddleware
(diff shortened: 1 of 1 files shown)


jeevatkm:
@jameshoulahan Thanks for your compliment and the PR contribution.

I agree, if dynamically the middleware is getting added/removed from resty then RWMutex is needed!

I just noticed `codecov` reports are not coming in, so currently, many PR are blocked. Let me figure out that, then I will merge this PR once PR validation is done.

jeevatkm:
@jameshoulahan I'm sorry for the late response from my end.

jeevatkm (changes requested):
@jameshoulahan My bad, I messed up the gofmt while resolving the code conflict with web editor. Could you please revise the PR?

FYI, I have fixed the codecov and PR validation issues.

jeevatkm (approved):
Thanks @jameshoulahan, I took care at my end.

PR #605, opened 2022-11-29
feat: Export transport method to return current transport from the client

### LINKED ISSUE #601: client Transport is not exported
The doc comment for suggests this function should be exported, but it is not:

```
// Transport method returns `*http.Transport` currently in use or error
// in case currently used `transport` is not a `*http.Transport`.
func (c *Client) transport() (*http.Transport, error) {
	if transport, ok := c.httpClient.Transport.(*http.Transport); ok {
		return transport, nil
	}
	return nil, errors.New("current transport is not an *http.Transport instance")
}
```

My use case here is I'm trying to integrate otelhttp, so I would like to do
```
client.SetTransport(otelhttp.NewTransport(client.Transport()))
````
to wrap the resty transport to inject the span context into the outbound request, but since the transport function is not exported, this won't work.

### PR DESCRIPTION
**Pull request by MicBun**
Fixes #601

### DIFF
--- a/client.go
+++ b/client.go
@@ -792,7 +792,7 @@ func (c *Client) SetRetryResetReaders(b bool) *Client {
 //
 // Note: This method overwrites existing `TLSClientConfig`.
 func (c *Client) SetTLSClientConfig(config *tls.Config) *Client {
-	transport, err := c.transport()
+	transport, err := c.Transport()
 	if err != nil {
 		c.log.Errorf("%v", err)
 		return c
@@ -809,7 +809,7 @@ func (c *Client) SetTLSClientConfig(config *tls.Config) *Client {
 //
 // Refer to godoc `http.ProxyFromEnvironment`.
 func (c *Client) SetProxy(proxyURL string) *Client {
-	transport, err := c.transport()
+	transport, err := c.Transport()
 	if err != nil {
 		c.log.Errorf("%v", err)
 		return c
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by xD-saleem — APPROVED** [review]

**Review by xD-saleem — DISMISSED** [review]

**Comment by MicBun** [comment]
Can this be merged or closed? I don't have write access

**Review by segevda — COMMENTED** [review]
Most of the lines changed in this commit are fixing the indentation. IMO you should remove these changes from this commit as they are unrelated. Another PR can be created to fix the indentation.

**Comment by MicBun** [comment]
> Most of the lines changed in this commit are fixing the indentation. IMO you should remove these changes from this commit as they are unrelated. Another PR can be created to fix the indentation.

thank you for the input

**Review by segev-indegy — APPROVED** [review]

**Comment by jeevatkm** [comment]
@MicBun Thanks for your PR and contribution. I'm sorry for the delayed attention on the PR.

May I know the context of the PR! I couldn't understand, why its needed.

**Review by jeevatkm — APPROVED** [review]
@MicBun Thanks for your contribution.

**Review by jeevatkm — CHANGES_REQUESTED** [review]
@MicBun I just noticed, this PR contains unnecessary commits. Can you please update your PR with only your changes or create a new?

**Comment by MicBun** [comment]
> @MicBun I just noticed, this PR contains unnecessary commits. Can you please update your PR with only your changes or create a new? After that we can merge this one.

sure, thank you for the feedback

**Comment by loris007** [comment]
fyi: the transport is already exported via http.Client
```
client := resty.New()
transport := client.GetClient().Transport
```

jeevatkm (approved):
Thanks @MicBun

MicBun:
> Thanks @MicBun

@jeevatkm You're welcome! I'm glad that I could contribute and be of help.

PR #619, opened 2023-02-23
Update: PreRequestHook use resty request struct for custom log

ifooth:


segevda (commented):
it seems like there was some automated linter running here, while this might be a positive thing, it clutters the commit, and should be done in a separate housekeeping commit. Can you please fix so that this will only contain the PreRequestHook change?

ifooth:
hi @segevda ,sorry for the late reply, let's review the design
- `OnBeforeRequest`:  change resty.Request some data, behave
- `SetPreRequestHook`: use for log or some else, diff with `OnBeforeRequest` is the had a http.Request init already
- OnAfterResponse:  use for log or done something for reponse
- OnError: use for log or done something for handler err, for example, connection err
- OnRequestLog: change header or body before log, use for remove useless or sensitive information
- OnResponseLog: change header or body before log, use for remove useless or sensitive information

my PR change SetPreRequestHook http.Request to resty.Request for custom log, because http.Request body is io.ReadCloser, read body is inefficient and unnecessary

jeevatkm:
@ifooth Thanks for your PR and contribution. 

* The `SetPreRequestHook` method is specifically designed with argument `http.Request` for Resty users, so that they could manipulate the RawRequest before request is fired from the Resty.
* I have read your [comment](https://github.com/go-resty/resty/pull/619#issuecomment-1454335723), I couldn't understand the rational behind changing argument. Also this PR is breaking change for existing Resty user who use `SetPreRequestHook` method. Typically breaking changes are meant for major version release/update.

Help me understand the rational for argument type change, currently I do not see a reason to merge this PR.

jeevatkm:
With above explanation, I'm closing this one.

PR #621, opened 2023-02-27
client: setters for JSON and XML Marshal/Unmarshal

GRbit:
Today I learned that I can set external JSON library for resty to marshal JSON faster. It wasn't so obvious because I haven't read the docs (let's be honest it's not only me who doesn't read docs), but there was no method for `resty.Client` to override JSON marshal function.

I suggest adding these `Set*` functions because this way they will be available through IDEs auto-completion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,13 +369,13 @@ import jsoniter "github.com/json-iterator/go"
 
 json := jsoniter.ConfigCompatibleWithStandardLibrary
 
-client := resty.New()
-client.JSONMarshal = json.Marshal
-client.JSONUnmarshal = json.Unmarshal
+client := resty.New().
+    SetJSONMarshaler(json.Marshal).
+    SetJSONUnmarshaler(json.Unmarshal)
 
 // similarly user could do for XML too with -
-client.XMLMarshal
-client.XMLUnmarshal
+client.SetXMLMarshaler(xml.Marshal).
+    SetXMLUnmarshaler(xml.Unmarshal)
 ```
 
 ### Multipart File(s) upload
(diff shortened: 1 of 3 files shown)


jeevatkm (changes requested):
@GRbit Thanks for your PR. Your PR description put a smile on my face, thanks.

Can you please test cases for this PR? [see here for more info](https://github.com/go-resty/resty/pull/621/checks?check_run_id=11780736846)

GRbit:
@jeevatkm I'm glad that it makes you smile.

Sorry for taking your time on that, but could you please tell me what kind of tests should I do? The only test I can come up with is something like this:
```
func TestClientJSON(t *testing.T) {
	m := func (v interface{}) ([]byte, error) { return nil,nil }
	c := New().SetJSONMarshaler(m)
	assertEqual(t, m, c.JSONMarshal)
}
```
But it looks silly for me to test such things because I don't think this test could possibly be helpful.

If this is what you looking for – I'll be glad to add a test for each function. If not, please show me some examples in the code

jeevatkm:
@GRbit Thanks for the getting back. 

> @jeevatkm I'm glad that it makes you smile.
> 
> Sorry for taking your time on that ...

Just to be clear, why I mentioned about smile is regarding this part from the PR description. Because, time-to-time, I also do the same! 
> (let's be honest it's not only me who doesn't read docs)

----

I understand, what you mean. However, based on the PR context, we are introducing a new setter methods for populating certain field. So it make sense to add a tiny test case which calls those setter(s) and does the assertion.

GRbit:
@jeevatkm I respect your standards, so my only question was "how should I make those tests?". Please check out the code I've just added, I hope it's ok.

jeevatkm (approved):
Thanks @GRbit for your contribution.

PR #625, opened 2023-03-09
Remove usages of deprecated io/ioutil

segevda:
io/ioutil has been deprecated since go1.16, and its functionality has been moved to io and os packages.


segevda:
I don't see how a commit that adds no new statements (quite the opposite - it removes several statements) can lower the coverage. @jeevatkm please take a look :pray:

PR #633, opened 2023-03-20
fix: client panic on missing digest value

segevda:
After splitting key=value pairs, we check that there are really key AND value. A maliciously crafted response header can cause the client to panic.


jeevatkm (approved):
@segevda Thanks for the PR.

PR #642, opened 2023-03-28
Access single request debug dump

softkot:
I do find the way to dump requests on a resty.Client level with SetDebug call but i did not find the simple way to control it on Request level. This pull request implement the Request level method SetOnRequestDump that can be used to access the dump on particular request regardless of the whole client debuging state.

```go 
request.SetOnRequestDump(func(r *resty.Request, dump string) {
			// TO DO  something with composed dump
		})
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -71,7 +71,7 @@ var (
 
 type (
 	// RequestDumpFunction is used for dump on request level instead on whole client level
-	RequestDumpFunction func(r *Request, sump string)
+	RequestDumpFunction func(r *Request, dump string)
 
 	// RequestMiddleware type is for request middleware, called before a request is sent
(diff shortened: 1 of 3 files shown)


jeevatkm (changes requested):
@softkot I'm wondering, instead adding this arbitrary function. I think it is best to add method `SetDebug(...)` at request level. It will align with Resty's design.  Could you update the PR?

softkot:
Try v2.9.1 and i can achive the required result with request.SetDebug + request.SetLogger with custom logger implementation.

This PR is outdated. Thanks.

jeevatkm:
@softkot Thanks for the feedback.

PR #654, opened 2023-05-05
Improve 204 no content handling

### LINKED ISSUE #652: Error object is not unset for 204 status code
Currently, if we set the object for automatic marshaling using [Request.SetError()](https://pkg.go.dev/github.com/go-resty/resty/v2#Request.SetError), it will not be unset in case of 204 status code. This could be problematic for error handling.

https://github.com/go-resty/resty/blob/15cedd4f9ee3df956be40f38fa9b85fb529c8078/middleware.go#L321-L324

### LINKED ISSUE #653: Incorrect request body empty check
Request body would no longer be `nil` (but rather an empty slice) after `ioutil.ReadAll` call. It would be more appropriate to do a length check here.

The current code would print an error due to JSON processing on an empty slice

https://github.com/go-resty/resty/blob/15cedd4f9ee3df956be40f38fa9b85fb529c8078/response.go#L158-L177

### LINKED ISSUE #678: Non nil body but empty in response cause 'unexpected end of JSON input' error
resty v2.7.0
go 1.20

@jeevatkm  I was implementing one client and the server is probably with issue in one of the routes that sends 201 and body is empty but r.body is getting initialized somehow although its length is 0. Maybe empty string is coming from server. 
It also affects `SetResult()`, maybe a check in `parseResponseBody` is also needed to prevent error from unmarshalling empty `[]byte{}`

if we change from:

```
func (r *Response) fmtBodyString(sl int64) string {
	if r.body != nil  {
		if int64(len(r.body)) > sl {
...
```
Actual: Here we can see there's nothing in body, maybe empty string? but Content-Type is application/json; charset=utf-8

Seleção_034

to:

```
func (r *Response) fmtBodyString(sl int64) string {
	if r.body != nil && len(r.body) > 0 {
		if int64(len(r.body)) > sl {
...
```
After this small change everything is fine now:

Seleção_033

haivp3010:
fixes #652
fixes #653
fixes #678

### DIFF
--- a/response.go
+++ b/response.go
@@ -93,7 +93,7 @@ func (r *Response) Cookies() []*http.Cookie {
 
 // String method returns the body of the server response as String.
 func (r *Response) String() string {
-	if r.body == nil {
+	if len(r.body) == 0 {
 		return ""
 	}
 	return strings.TrimSpace(string(r.body))
@@ -156,7 +156,7 @@ func (r *Response) setReceivedAt() {
 }
 
 func (r *Response) fmtBodyString(sl int64) string {
-	if r.body != nil {
+	if len(r.body) > 0 {
 		if int64(len(r.body)) > sl {
 			return fmt.Sprintf("***** RESPONSE TOO LARGE (size - %d) *****", len(r.body))
 		}
--- a/middleware.go
+++ b/middleware.go
@@ -320,6 +320,7 @@ func responseLogger(c *Client, res *Response) error {
 
 func parseResponseBody(c *Client, res *Response) (err error) {
 	if res.StatusCode() == http.StatusNoContent {
+		res.Request.Error = nil
 		return
 	}
 	// Handles only JSON or XML content type


response.go
```diff
@@ -93,7 +93,7 @@ func (r *Response) Cookies() []*http.Cookie {
 
 // String method returns the body of the server response as String.
 func (r *Response) String() string {
-	if r.body == nil {
+	if len(r.body) == 0 {
```
jeevatkm, line 96: @haivp3010 Can you update this line as -
```go
if r.body == nil  || len(r.body) == 0 {
```

response.go
```diff
@@ -156,7 +156,7 @@ func (r *Response) setReceivedAt() {
 }
 
 func (r *Response) fmtBodyString(sl int64) string {
-	if r.body != nil {
+	if len(r.body) > 0 {
```
jeevatkm, line 159: @haivp3010 Can you update this line as -
```go
if r.body != nil && len(r.body) > 0 {
```

jeevatkm (changes requested):
@haivp3010 Thanks for your PR, appreciated.  

Please see the requested changes, also added possible test cases.

response.go
haivp3010: @jeevatkm I think it makes no difference here. Whether `r.body` is `nil` or empty, calling `len` on it would still return 0.

https://pkg.go.dev/builtin#len

```
The len built-in function returns the length of v, according to its type:

Array: the number of elements in v. Pointer to array: the number of elements in *v (even if v is nil). Slice, or map: the number of elements in v; if v is nil, len(v) is zero. String: the number of bytes in v. Channel: the number of elements queued (unread) in the channel buffer; if v is nil, len(v) is zero.
```

response.go
haivp3010: @jeevatkm same as above

jeevatkm (approved):

PR #664, opened 2023-05-16
Support RawPathParams without escaping

### LINKED ISSUE #663: there is no way to use PathParams without escaping
### Discussed in https://github.com/go-resty/resty/discussions/570

<div type='discussions-op-text'>

<sup>Originally posted by **SVilgelm** August  3, 2022</sup>
Hi, as it is said in the subject. I want to use the PathParams without escaping. Is there anyway to do that?</div>

In our case the object Ids can include the `/` symbols, like `foo/bar/xyz`:
```go
resty.New().R().SetPathParam("id", "foo/bar/xyz").Get("https://example.com/{id}")
```

the composed url will be `https://example.com/foo%2Fbar%2Fxyz` instead of `https://example.com/foo/bar/xyz`

As a solution i would like to propose an additional `RawPathParams` option on client and request levels

SVilgelm:
Added `RawPathParams` options to `Client` and `Request` objects to support the path parameters with special characters, like `/`, without escaping.

Fix #663
Closes #570

### DIFF
--- a/client.go
+++ b/client.go
@@ -103,6 +103,7 @@ type Client struct {
 	QueryParam            url.Values
 	FormData              url.Values
 	PathParams            map[string]string
+	RawPathParams         map[string]string
 	Header                http.Header
 	UserInfo              *User
 	Token                 string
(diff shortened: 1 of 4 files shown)


SVilgelm:
Hi @jeevatkm, could you please review this PR?

SVilgelm:
Hi @jeevatkm, is there a way to get your attention to review this PR?

jeevatkm (approved):
@SVilgelm Thanks for your PR, appreciated.

PR #674, opened 2023-06-27
fix: parseResponseBody overrides original error code in case of unmar…

### LINKED ISSUE #673: parseResponseBody overrides original error code in case of unmarshalling error
If the client has an `Error` object for automatic unmarshalling for the request, and the JSON response cannot be unmarshalled, the current behavior is to override the HTTP response status with a marshalling error like `unexpected end of JSON input`.

Since we cannot lose this information, I suggest to change:
```
                        if res.Request.Error != nil {
                              err = Unmarshalc(c, ct, res.body, res.Request.Error)
                        }

```
to:
```
                        if res.Request.Error != nil {
                              unmarshalErr := Unmarshalc(c, ct, res.body, res.Request.Error)
                              if unmarshalErr != nil {
                                      log.Printf("Failed to unmarshal response body: %s", unmarshalErr)
                              }
                        }
```
This way we'll miss the unmarshalling error but at least we can handle the server response status.

Maybe there's a better way to report the error, but I haven't found anyone ATM, please suggest an alternative if any.

dmartinol:
This closes #673

I'm not sure what is the best practice to notify of the swallowed marshalling error, anyway I think we cannot override the original error received from the server because of that. Please let me know your feedback about this fix.

**Note**: I found this issue while using [gocloak](https://github.com/Nerzal/gocloak) the go client for Keycloak. Un fact, the latest versions of Keycloak have changed the response data for the 404/Not Found responses, and the body is empty instead of containing a JSON document, so the get requests fail with this `unexpected end of JSON input` error hiding the 404 status code, which is not acceptable.


renathoaz:
@dmartinol if you could do this small change too I will appreciate. Thank you in advance
https://github.com/go-resty/resty/issues/673#issuecomment-1644390362

middleware.go
```diff
@@ -342,7 +343,10 @@ func parseResponseBody(c *Client, res *Response) (err error) {
 			}
 
 			if res.Request.Error != nil {
-				err = Unmarshalc(c, ct, res.body, res.Request.Error)
+				unmarshalErr := Unmarshalc(c, ct, res.body, res.Request.Error)
+				if unmarshalErr != nil {
+					log.Printf("[WARN] Cannot unmarshal response body: %s", unmarshalErr)
```
jeevatkm, line 348: @dmartinol can you update this line as following?
```go
c.log.Warnf("Cannot unmarshal response body: %s", unmarshalErr)
```

jeevatkm (changes requested):

jeevatkm (approved):
Thanks, @dmartinol, for updating the PR.

PR #683, opened 2023-07-28
improve: support js,wasm

### LINKED ISSUE #610: Protocol not available when updating from go 1.17 to 1.19
I'm running resty in a wam application

I got this error with default example : 

```
Error: Get "http://httpbin.org/get": dial tcp: lookup httpbin.org: Protocol not available
wasm_exec.js:19 Response Status Code: 0
wasm_exec.js:19 Response Status: 
wasm_exec.js:19 Response Body: 
wasm_exec.js:19 Response Time: 400.128µs
```

https://gitlab.com/c1560/cryptofiscafacile/-/merge_requests/824

I have tested http.get successfully

shynome:
those code copy from `/go/src/net/http/transport_default_other.go`

closes #610

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -25,7 +25,7 @@ func createTransport(localAddr net.Addr) *http.Transport {
 	}
 	return &http.Transport{
 		Proxy:                 http.ProxyFromEnvironment,
-		DialContext:           dialer.DialContext,
+		DialContext:           transportDialContext(dialer),
 		ForceAttemptHTTP2:     true,
 		MaxIdleConns:          100,
 		IdleConnTimeout:       90 * time.Second,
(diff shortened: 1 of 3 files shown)


jeevatkm (changes requested):
@shynome Can you please add test case?

shynome:
I'm sorry, I don't have time to add test case because I don't use resty anymore.
now I just use http client with some function wrap

as you can see, those code is copy from golang std library `net/http/transport_default_other.go`

jeevatkm:
@shynome Thanks for getting back. I see; I will take it up later on.

jeevatkm:
Ref: https://github.com/golang/go/blob/master/src/net/http/transport_default_wasm.go

jeevatkm (approved):

Issue #686, opened 2023-08-11
OnBeforeRequest and default middleware order documentation and code mismatch


leejuyuu:
The documentation of `OnBeforeRequest` 

https://github.com/go-resty/resty/blob/3d08b3602d13ae8a44f246db2ec34402e18497c6/client.go#L403-L405

says that the middleware set by this function is applied after the default middlewares.

However,  when the request is executed, the order is reversed.
https://github.com/go-resty/resty/blob/3d08b3602d13ae8a44f246db2ec34402e18497c6/client.go#L896-L912

I think this resulted in https://github.com/go-resty/resty/issues/517.

I am not sure which side is the intended result.

jeevatkm:
@leejuyuu Thanks for bringing it up. I didn't notice it. The documentation needs correction. I will correct it, or you could submit the PR. Please let me know.

As @0140454 mentioned in the https://github.com/go-resty/resty/issues/517#issuecomment-1120139445, this is the correct order of execution.

leejuyuu:
@jeevatkm Thanks for the reply! Opened #701.

PR #701, opened 2023-09-23
docs: fix OnBeforeRequst doc application order

leejuyuu:
The documentation about the order of middleware application for OnBeforeRequet was different from the implementation. Fixed this by changing the documentation.

Fixes https://github.com/go-resty/resty/issues/686.


jeevatkm (approved):
@leejuyuu Thanks for your PR to correct the docs.

PR #705, opened 2023-09-24
fix: correct url path params processing order #519

### LINKED ISSUE #519: the applying order of PathParams
Why first applying PathParams from Request and then Client(in function parseRequestURL). While in other situations, the applying order will be first Client, then Request.

jeevatkm:
Closes #519

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -27,25 +27,25 @@ const debugRequestLogKey = "__restyDebugRequestLog"
 
 func parseRequestURL(c *Client, r *Request) error {
 	// GitHub #103 Path Params
-	if len(r.PathParams) > 0 {
-		for p, v := range r.PathParams {
-			r.URL = strings.Replace(r.URL, "{"+p+"}", url.PathEscape(v), -1)
-		}
-	}
 	if len(c.PathParams) > 0 {
 		for p, v := range c.PathParams {
 			r.URL = strings.Replace(r.URL, "{"+p+"}", url.PathEscape(v), -1)
 		}
 	}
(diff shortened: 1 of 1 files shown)


SVilgelm:
@jeevatkm, the previous implementation was correct.
Let's say we have url `example.com/{foo}/{bar}/`
And have the parameters:
```
Client: {"foo": "123"}
Request: {"foo": "456", "bar": "789"}

r.URL = "example.com/{foo}/{bar}/"
```


the current code will apply the parameters from `Client` first:
* `"foo"`: 
```
r.URL = r.URL = strings.Replace(r.URL, "{foo}", url.PathEscape("123"), -1)`
---
r.URL = "example.com/123/{bar}/"
```
then the request parameters:
* `"foo"`:
```
r.URL = r.URL = strings.Replace(r.URL, "{foo}", url.PathEscape("456"), -1)`
---
r.URL = "example.com/123/{bar}/"
```
no changes because there is no `{foo}` placeholder in the `r.URL`
* `"bar"`:
```
r.URL = r.URL = strings.Replace(r.URL, "{bar}", url.PathEscape("789"), -1)`
---
r.URL = "example.com/123/789/"
```


so the request parameters must be applied first, then the client, or we need to change the logic to use a map and `strings.Replacer`

jeevatkm:
@SVilgelm Thanks for reminding me. Damn it, last night, how did I forget this!! 🤦 I will revert the changes.

SVilgelm:
We need to add the unit tests to prevent such thing in the future, I can do it later today

PR #710, opened 2023-09-25
fix TestClientOnResponseError parallel tests

SVilgelm:
In case of using `t.Parallel()` inside a for loop, the loop variables must be copied, otherwise only the last test will be executed.

```go
	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			t.Parallel()
```
changed to
```go
	for _, test := range tests {
		test := test
		t.Run(test.name, func(t *testing.T) {
			t.Parallel()
```

also the AuthServer must be created within the test, otherwise some tests fail.

`t.Log(test.Name)` can be used to verify the issue:
```go
	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			t.Parallel()
			t.Log("Test name:", test.name)
```
Screenshot 2023-09-24 at 8 22 42 PM

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -780,9 +780,6 @@ func TestNewWithLocalAddr(t *testing.T) {
 }
 
 func TestClientOnResponseError(t *testing.T) {
-	ts := createAuthServer(t)
-	defer ts.Close()
-
 	tests := []struct {
 		name        string
 		setup       func(*Client)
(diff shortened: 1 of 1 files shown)


jeevatkm (approved):
@SVilgelm Thanks for the unit test improvements.

SVilgelm:
I saw this issue so many times :)

jeevatkm:
It would be helpful if `go test` throws some warning or error if detects this case.

PR #717, opened 2023-10-01
Fix#383: initial change to make SetHeader() thread safe

amarjeetanandsingh:


amarjeetanandsingh:
Not only the `SetHeader()`, but we may need to consider other APIs too.

jeevatkm:
@amarjeetanandsingh, I appreciate your interest in creating PR. 

This PR is not a solution. A complete resty client object should be made thread-safe with global read-write Mutex. This approach will result in a few breaking changes, which I'm parking for v3.0.0 development. So, I will close this PR.

PR #734, opened 2023-10-16
example: add examples about ssl, debug, curl, trace, error, etc  && new feature about getting curl command

ahuigo:
## New feature: generate curl command
Example: https://github.com/ahuigo/resty/blob/examples-v2/examples/debug_curl_test.go

```
var curlCmdExecuted string
client := resty.New()
_, err := client.R().
    SetResultCurlCmd(&curlCmdExecuted).
    Get("https://httpbin.org/get")

// Explore curl command
fmt.Println("Curl Command:", curlCmdExecuted)
```

## Some examples
Test these examples: https://github.com/ahuigo/resty/blob/examples-v2/examples

```
$ go test -failfast  ./examples
ok      github.com/go-resty/resty/v2/examples   (cached)
```


jeevatkm:
@ahuigo, I'm putting my effort into v3. I will return to this PR after the v3 work is in good shape. Thanks.

PR #735, opened 2023-10-17
fix: Improve Digest WWW-Authenticate response header parsing compatibility

bearki:
Improve Digest WWW-Authenticate response header parsing compatibility

1. Use commas to disassemble WWW-Authenticate instead of commas and spaces.
2. Clear the spaces before and after each key-value pair.
3. Added double quote clearing operation for stale and algorithm values.


jeevatkm (approved):
@bearki Thanks for the PR, appreciated.

Issue #739, opened 2023-10-25
DATA RACE in resty's middleware (go-resty/resty/v2.releaseBuffer())


1713266:
We detected a data race in resty's middleware when releasing/resetting the buffer. Please check.

Go Version: 1.21.1
Resty Version: v2.10.0

```
==================
WARNING: DATA RACE
Read at 0x00c001e91ce0 by goroutine 537694:
  bytes.(*Buffer).Reset()
      /home/fwkadmin/.ift/import/content/go-1.21.1-bin-linuxx86_64/go/src/bytes/buffer.go:107 +0x46
  github.com/go-resty/resty/v2.releaseBuffer()
      /testDirectory/vendor/github.com/go-resty/resty/v2/util.go:298 +0x41
  github.com/go-resty/resty/v2.handleRequestBody()
      /testDirectory/vendor/github.com/go-resty/resty/v2/middleware.go:484 +0x6e
  github.com/go-resty/resty/v2.parseRequestBody()
      /testDirectory/vendor/github.com/go-resty/resty/v2/middleware.go:193 +0x290
  github.com/go-resty/resty/v2.(*Client).execute()
      /testDirectory/vendor/github.com/go-resty/resty/v2/client.go:1164 +0x550
  github.com/go-resty/resty/v2.(*Request).Execute.func2()
     /testDirectory/vendor/github.com/go-resty/resty/v2/request.go:947 +0x179
  github.com/go-resty/resty/v2.Backoff()
     /testDirectory/vendor/github.com/go-resty/resty/v2/retry.go:115 +0x1d5
  github.com/go-resty/resty/v2.(*Request).Execute()
      /testDirectory/vendor/github.com/go-resty/resty/v2/request.go:941 +0xac4
  github.com/go-resty/resty/v2.(*Request).Post()
      /testDirectory/vendor/github.com/go-resty/resty/v2/request.go:862 +0x344
  github.wdf.sap.corp/ift/sl-core/pkg/slp/client.sumHandler.sendUserInput()
      /testDirectory/vendor/github.wdf.sap.corp/ift/sl-core/pkg/slp/client/sum_handler.go:69 +0x2cf
  github.wdf.sap.corp/ift/sl-core/pkg/slp/client.(*sumHandler).sendUserInput()
      <autogenerated>:1 +0x84
  github.wdf.sap.corp/ift/sl-core/pkg/slp/client.(*Client).SendUserInput()
[...]

PR #745, opened 2023-11-02
Security: Don't put the same bytes.Buffer into sync.Pool twice

### LINKED ISSUE #743: Request body is written multiple times
In v2.10.0, I've observed an issue where the request body passed to `.SetBody(...)` is intermittently written multiple times under certain conditions.

From what I gather, it is somehow related to the retry mechanism, or at the very least I was only able to reproduce by triggering that.

[In the repro here](https://github.com/patrikeh/resty-request-body-issue), I ping an httptest mock endpoint triggering the retry mechanism, and observe the JSON request body being written multiple consecutive times e.g. 
```
{ <some-json> }{ <some-json> }
```

This issue does not occur in v2.9.1.

lattwood:
This fixes #743, possibly #739.

Discovered this while tracking down an issue in https://github.com/linode/terraform-provider-linode.

The removed call to sync.Pool.Put is handled by the request body wrapper that puts the buffer back in the pool when the body is closed.

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -481,7 +481,6 @@ func handleContentType(c *Client, r *Request) {
 
 func handleRequestBody(c *Client, r *Request) error {
 	var bodyBytes []byte
-	releaseBuffer(r.bodyBuf)
 	r.bodyBuf = nil
 
 	switch body := r.Body.(type) {


lattwood:
This closes #739 as well.

Now that I’m thinking about it, this is a security issue, as it’s disclosing a HTTP body to an unrelated endpoints/servers.

fabiomargarido:
Hey @lattwood, this seems like a simple enough fix and it's been available here for almost a month now. Is anything still missing before this can be merged and the security issue mitigated?
Thanks.

reedloden:
@jeevatkm Able to take a look here when you have a moment?

safaci2000:
Any updates on this ticket?

lattwood:
Your best bet is going to be forking this, or moving to a different library, based on how long this issue has been open.

niksteff:
Yeah we are moving. Not only is the CVE a big problem, after the update to 2.10 we experienced fun bugs where the json request body of a previous or the same request (could not tell) was copied in the host part of the request URL. This occurred when multiple request were running concurrently and we changed nothing but the resty upgrade. Was fine for two years. I cannot prove it as I could not recreate the bug but a downgrade to 2.9.1 immediately solved the problem.

It could be a side effect of this buffer problem but as i cannot prove it i won't open a new thread. Just wanted to state it if somebody else experiences something different and has more time at hands to test it through.

Here is an example log from our kibana:

```json
2023/11/28 13:51:16.736214 WARN RESTY parse "{"some very large json request body here"}/oms/customer/v3/users/2133478": first path segment in URL cannot contain colon, Attempt 1
```

Because of the `WARN RESTY` you can see the warning is logged from the resty code base as this log is hard coded there.

Coronon:
For everyone that needs a quick fix without migrating away from go-resty:

```go-module
replace github.com/go-resty/resty/v2 => github.com/lattwood/resty/v2 v2.0.0-20231102200459-74e9e135ae0c
```

vithubati:
Any time line as to when this PR will be merged?

jeevatkm:
@lattwood and members on the PR - Thanks for reaching out. I have been traveling on vacation days and will return from vacation in the first week of January.
I'm sorry for not checking my emails and notifications properly these days.

Let me merge this PR and make a release.

jeevatkm (approved):
@lattwood Thank you for the PR, appreciated.

PR #774, opened 2024-02-10
Add Client.Clone function

### LINKED ISSUE #773: Clone or copy of existing client
I have a use case where my application should have a global HTTP client with sane defaults for tracing, logging and error handling and then some submodules which must apply highly specific settings for only their use case.

For that I would like to copy / clone an existing client which thus inherits all of the clients settings. In other libraries / languages this is also sometimes possible by having a child-parent relationship between clients.

What would be the best way to approach this with resty? Is there a function or common solutions for this use case?

Edit: If this is *not* a present feature but of interest to people, let's discuss how that could be implemented.

fabiante:
This adds a crude cloning function. It can be used to have a main client with default values and then create copies from that, which must use custom values.
    
The added function is rather crude and may introduce the potential that, when adding a new Client field, we forget to add it to Clone.

Thus, we should discuss if this is the right way to go. Using code generation or relflection might be a more maintainable solution.

Merging this PR would close #773

### DIFF
--- a/client.go
+++ b/client.go
@@ -1125,6 +1125,25 @@ func (c *Client) GetClient() *http.Client {
 	return c.httpClient
 }
 
+// Clone returns a clone of the original client.
+//
+// Be carefull when using this function:
+// - Interface values are not deeply cloned. Thus, both the original and the clone will use the
+// same value.
+// - This function is not safe for concurrent use. You should only use this when you are sure that
+// the client is not being used by any other goroutine.
+//
+// Since v2.12.0
+func (c *Client) Clone() *Client {
+	// dereference the pointer and copy the value
+	cc := *c
+
+	// lock values should not be copied - thus new values are used.
+	cc.afterResponseLock = &sync.RWMutex{}
+	cc.udBeforeRequestLock = &sync.RWMutex{}
+	return &cc
+}
+
 //‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
 // Client Unexported methods
 //_______________________________________________________________________
--- a/client.go
+++ b/client.go
@@ -142,11 +142,11 @@ type Client struct {
 	proxyURL            *url.URL
 	beforeRequest       []RequestMiddleware
 	udBeforeRequest     []RequestMiddleware
-	udBeforeRequestLock sync.RWMutex
+	udBeforeRequestLock *sync.RWMutex
 	preReqHook          PreRequestHook
 	successHooks        []SuccessHook
 	afterResponse       []ResponseMiddleware
-	afterResponseLock   sync.RWMutex
+	afterResponseLock   *sync.RWMutex
 	requestLog          RequestLogCallback
 	responseLog         ResponseLogCallback
 	errorHooks          []ErrorHook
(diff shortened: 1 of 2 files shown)


jeevatkm (changes requested):
@fabiante Thanks for creating a draft.  I have read the issue thread #773.

Can you please update the PR as follows?

Add a new method
```go
// Clone returns a clone of the original client.
//  <add further doc info here>
//
// Since v2.12.0
func (c *Client) Clone() *Client {
	// dereference the pointer and copy the value
	cc := *c
	return &cc
}
```

Change these field definitions to a pointer type.
* https://github.com/go-resty/resty/blob/97187c431cf6db12100c49b68dab25bca79098c1/client.go#L145
* https://github.com/go-resty/resty/blob/97187c431cf6db12100c49b68dab25bca79098c1/client.go#L149
```go
udBeforeRequestLock *sync.RWMutex
.
.
.
afterResponseLock   *sync.RWMutex
```

Finally, initialize the fields while creating a new client
https://github.com/go-resty/resty/blob/97187c431cf6db12100c49b68dab25bca79098c1/client.go#L1343

client.go
```diff
@@ -1125,6 +1125,75 @@ func (c *Client) GetClient() *http.Client {
...
+// Clone returns a clone of the original client.
+//
+// Be carefull when using this function:
+// - Interface values are not deeply cloned. Thus, both the original and the clone will use the
+// same value.
+// - This function is not safe for concurrent use. You should only use this when you are sure that
+// the client is not being used by any other goroutine.
+//
+// Since v2.12.0
+func (c *Client) Clone() *Client {
```
jeevatkm, line 1137: @fabiante Thanks for updating the PR. I believe I requested this method replacement with -
```go
func (c *Client) Clone() *Client {
	// dereference the pointer and copy the value
	cc := *c
	return &cc
}
```

I still see the previous code. Did you face any issues?

jeevatkm (changes requested):

client.go
fabiante: Oh right. Id skipped over that part I guess. I will look into that.

fabiante: I originally intended to implement the copy function like this. I did **not** do it because the linter pointed out that the dereferencing would copy the mutex value: `assignment copies lock value`

With the fields changed to pointers, this warning is gone. But is this really okay?

AFAIK there will still be only one mutext which is now shared across clients. Is that the way we want to go? Fine by me but I think I don't yet have a deep understanding of how this could effect users.

I will push the change in case we are fine with that approach.

jeevatkm: @fabiante I will validate and get back

client.go
```diff
@@ -1125,6 +1125,21 @@ func (c *Client) GetClient() *http.Client {
...
+	// dereference the pointer and copy the value
+	cc := *c
+	return &cc
```
jeevatkm, line 1140: @fabiante Just a while ago, I validated it. We need to reinitialize the lock after the clone.

The complete clone method is as follows. Can you please update the PR?
```go
func (c *Client) Clone() *Client {
	// dereference the pointer and copy the value
	cc := *c
	cc.afterResponseLock = &sync.RWMutex{}
	cc.udBeforeRequestLock = &sync.RWMutex{}
	return &cc
}
```

jeevatkm (changes requested):

client.go
fabiante: I also think this makes more sense. Linters my still give a warning / hint about the dereferencing but the code should obiously show that we deal with this appropriately.

client.go
fabiante: Resolving since your latest comment adresses the same topic.

Issue #781, opened 2024-03-05
Path param encoding is broken


sakateka:
Hi!
I have the following code example.
```go
c.SetBaseURL("https://example.com/prefix")
r.SetPathParam("first", "1").
        SetPathParam("second", "2")
r.Get("{first}/{second}")
```
And after the release v2.10.0-rc.1 this example started failing.
The debug logs show the following
```
GET /prefix/%7Bfirst%7D/%7Bsecond%7D
HOST   :  example.com
```

I wrote the test and corrected the code, the PR is attached.

PR #782, opened 2024-03-05
fix: encode path params with BaseURL and the first param at index zero (#781)

sakateka:
Related issue https://github.com/go-resty/resty/issues/781


jeevatkm (approved):
@sakateka Thanks for the PR and fixing the big.

PR #785, opened 2024-04-01
build: update github actions

segevda:
actions/checkout@v4
actions/setup-go@v5

both have faster runtime (node 20) and bugfixes


jeevatkm (approved):
Thanks for the PR @segevda

PR #789, opened 2024-04-12
Fix digest auth

phw:
This re-implements parsing the digest challenge to fix authentication against servers returning multiple values for `qop`.

Take this example from https://httpwg.org/specs/rfc7616.html:

```
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Digest
    realm="http-auth@example.org",
    qop="auth, auth-int",
    algorithm=SHA-256,
    nonce="7ypf/xlj9XXwfDPEoM4URrv/xwf94BcCAzFZH4GiTo0v",
    opaque="FQhe/qaU925kfnzjCev0ciny7QMkPqMAFRtzCUYo5tdS"
```

Here qop is set to `qop="auth, auth-int"`. Because the previous parser split the entire list by comma to obtain the key value pairs this gets separated into `qop="auth` and ` auth-int"`, which eventually failes to parse.

The new parser goes over the challenge rune by rune and does not split when inside a quotation.

Also some servers will respond with `qop="auth,auth-int"` (no space after the comma). Hence also adjust `validateQop` to handle this.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -104,6 +104,7 @@ func TestClientDigestAuth(t *testing.T) {
 func TestClientDigestSession(t *testing.T) {
 	conf := defaultDigestServerConf()
 	conf.algo = "MD5-sess"
+	conf.qop = "auth, auth-int"
 	ts := createDigestServer(t, conf)
 	defer ts.Close()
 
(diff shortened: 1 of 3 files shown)


jeevatkm (approved):
@phw Thanks for the PR.

PR #817, opened 2024-08-09
doc(readme): Add a note, GenerateCurlCommand needs to turn on EnableT…

ahuigo:
Since the body cannot be found in the curl command, some friends think that `SetBody` does not take effect.

In order to avoid this, I add a note in readme to remind everyone that **GenerateCurlCommand needs to turn on EnableTrace**


jeevatkm (approved):
@ahuigo Thanks for the PR

PR #820, opened 2024-08-13
feat: add ability to set custom multipart boundary value

PokeGuys:
### Use case

Integrating with some web service that require a specific boundary format in multipart/form-data requests.

Some legacy service might requires a specific prefix in the boundary value, to ensure the compatibility, we need the ability to set a custom boundary value in the multipart request.

### Example

```go
package main

import (
	"crypto/rand"
	"fmt"
	"io"
	"log"

)

const BoundaryPrefix = "Some_MultiPart_Prefix_"

func randomBoundary() string {
	var buf [4]byte
	_, err := io.ReadFull(rand.Reader, buf[:])
	if err != nil {
		panic(err)
	}
	return fmt.Sprintf("%s_%X", BoundaryPrefix, buf[:])
}

func main() {
    // Initialize the client
    client := resty.New()

    // Perform the file upload
    _, err := client.R().
        SetMultipartFormData(map[string]string{"test": "1"}).
        SetBoundary(randomBoundary()) // Setting the custom boundary
        Post("http://example.com/upload")
    if err != nil {
        log.Fatalf("Failed to upload file: %v", err)
    }
}
```

### DIFF
--- a/request.go
+++ b/request.go
@@ -69,6 +69,7 @@ type Request struct {
 	bodyBuf             *bytes.Buffer
 	clientTrace         *clientTrace
 	log                 Logger
+	multipartBoundary   string
 	multipartFiles      []*File
 	multipartFields     []*MultipartField
 	retryConditions     []RetryConditionFunc
@@ -455,6 +456,13 @@ func (r *Request) SetMultipartField(param, fileName, contentType string, reader 
 func (r *Request) SetMultipartFields(fields ...*MultipartField) *Request {
 	r.isMultiPart = true
 	r.multipartFields = append(r.multipartFields, fields...)
+	return r
+}
+
+// SetBoundary method sets the multipart boundary for the request
+// By default a random boundary will be generated in `mime/multipart`.
+func (r *Request) SetBoundary(boundary string) *Request {
+	r.multipartBoundary = boundary
 	return r
 }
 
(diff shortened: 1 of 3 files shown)


jeevatkm:
@PokeGuys - Can you please add one practical use case and its example in the PR description?

PokeGuys:
@jeevatkm Hi. Updated.

request.go
```diff
@@ -458,6 +459,13 @@ func (r *Request) SetMultipartFields(fields ...*MultipartField) *Request {
 	return r
 }
 
+// SetBoundary method sets the multipart boundary for the request
+// By default a random boundary will be generated in `mime/multipart`.
+func (r *Request) SetBoundary(boundary string) *Request {
```
jeevatkm, line 464: @PokeGuys Thanks for adding the use case detail to the PR description.

Can you refactor this method as below and it includes version info too?
```go
// SetMultipartBoundary method sets the custom multipart boundary for the multipart request.
// Typically, the `mime/multipart` package generates a random multipart boundary, if not provided.
//
// Since v2.15.0
func (r *Request) SetMultipartBoundary(boundary string) *Request {
 	r.multipartBoundary = boundary
 	return r
 }
```

jeevatkm (changes requested):

request.go
PokeGuys: Updated.

jeevatkm (approved):
Thanks, @PokeGuys, for the PR.

PR #844, opened 2024-09-03
fix: buildCurlRequest() replace req.Body with req.GetBody()

tttturtle-russ:
```The `TestGenerateExecutedCurl` failure started after I merged from branch `v2` to branch `main`. It needs to investigate why it's failing.```

_Originally posted by @jeevatkm in https://github.com/go-resty/resty/issues/827#issuecomment-2323726014_

Hi jeevatkm, I've found the reason why the test case failed.
In https://github.com/go-resty/resty/blob/6d941ac6f1ed35557b6fb3fb41cf74beeb9be2f0/client.go#L1601 `Do` method will read the `req.RawRequest.Body` and make it empty. When generating curl body it will read an empty string, and that's why it failed.

This commit offers a temporary solution to this problem, I think the ultimate solution is to add a `Clone` method in `Request`, as #718 describes.

### DIFF
--- a/util_curl.go
+++ b/util_curl.go
@@ -31,7 +31,12 @@ func buildCurlRequest(req *http.Request, httpCookiejar http.CookieJar) (curl str
 
 	// 3. Generate curl body
 	if req.Body != nil {
-		buf, _ := io.ReadAll(req.Body)
+		// httpclient.Do method will read the entire body and make it empty
+		// thus using req.GetBody() instead of req.Body since req.Body
+		// is empty after Do method.
+		// body here is a copy of original body
+		body, _ := req.GetBody()
+		buf, _ := io.ReadAll(body)
 		req.Body = io.NopCloser(bytes.NewBuffer(buf)) // important!!
 		curl += `-d ` + shellescape.Quote(string(buf))
 	}

### CODE REVIEW

File: util_curl.go [hunk 1]
```diff
@@ -31,7 +31,12 @@ func buildCurlRequest(req *http.Request, httpCookiejar http.CookieJar) (curl str
 
 	// 3. Generate curl body
 	if req.Body != nil {
-		buf, _ := io.ReadAll(req.Body)
+		// httpclient.Do method will read the entire body and make it empty
+		// thus using req.GetBody() instead of req.Body since req.Body
+		// is empty after Do method.
+		// body here is a copy of original body
+		body, _ := req.GetBody()
+		buf, _ := io.ReadAll(body)
```
jeevatkm, line 39: @tttturtle-russ Thanks for creating a PR. Last night, I discovered the branch `main` was missing curl middleware. It seems it happened when I merged the branch `v2` into branch `main`.

Branch `main`
https://github.com/go-resty/resty/blob/6d941ac6f1ed35557b6fb3fb41cf74beeb9be2f0/resty.go#L195-L196

Branch `v2`
https://github.com/go-resty/resty/blob/370d744bf0a478e71832d591061b30a9f91ef402/client.go#L1484-L1486

jeevatkm (commented):
@tttturtle-russ Let's decline this PR and create a new PR with adding middleware would solve. Also, while on PR #842, I noticed, the CURL follow needs an improvement I will work on later.

tttturtle-russ:
@jeevatkm OK, I will close this PR and create a new one with curl middleware added.

Issue #856, opened 2024-09-18
Feature request : Implement ability to dynamically reload TLS certificates


cdoucy:
### Description

`client.SetRootCertificate` and `client.SetClientRootCertificate` allow to configure certificates for TLS communication.
It would be great if it was possible to dynamically reload certificates when they expire to automatically handle certificate rotation.

### Use-case

Assume there are multiple HTTPS services in a Kubernetes cluster. These HTTPS services are using Certificates issued by [cert-manager](https://github.com/cert-manager/cert-manager).
Therefore, pods who host HTTPS client need to be mounted with the root certificate. When `cert-manager` rotate the Certificates, the running HTTPS clients need to reload the mounted certificate.

To work around it, it's possible to either panic on TLS errors to restart the pod and reload the cert, or to create a fresh new `resty` Client for each HTTPS call, but that's not very efficient.

### Possible implementations

I see two ways to implement this feature:

1. Watch for file system events and reload the certificate when the file changes

[fsnotify](https://github.com/fsnotify/fsnotify) is a cross-platform library that allows to watch file system events. I believe it would be a good fit for the use-case.
However, I'm not sure if `fsnotify` can be used in `resty` due to legal reasons. (`fsnotify` is BSD licensed while `resty` is MIT)
Implementing a cross-platform module to watch FS system would require some engineering efforts.

2. Poll periodically certificate file mod time and reload it when it changes

A routine could periodically call `os.Stat()` on the file to check the certificate file modification time.
This approach would be straightforward to implement but is not as efficient as the previous option.

cdoucy:
Hello @jeevatkm !
I would happy to start contributing to `resty` by addressing this issue.

jeevatkm:
Hello @cdoucy - Thanks for reaching out and proposing a feature request. I agree; it is a helpful feature for the Resty users.

I suggest a change in the implementation. What do you think of the following approach?

> Add functionality with user-provided `time.Duration`, and Resty to reload the certificates when the timer hits by checking `os.Stat` instead of polling.

Please feel free to create a PR.

_PS: Resty v2 Client settings modification is not set thread-safe 100%. It should be all right if there are no concurrent certificate modifications; it is just not immune to it yet. The upcoming Resty v3 Client comes with `sync.RWMutex` to manage settings on client instances, which is 100% thread-safe._

jeevatkm:
@cdoucy I edited the above suggestion with input parameter details.

cdoucy:
[...]

PR #866, opened 2024-09-22
feat: Implement ability to dynamically reload Root & Client Root TLS certificates

cdoucy:
Fixes https://github.com/go-resty/resty/issues/856


jeevatkm (changes requested):
@cdoucy Thanks for creating a PR. Please take a look at my feedback. 

* Make use of https://pkg.go.dev/time#Ticker to execute the function to reload based on a timer
* Do not use `OnBeforeRequest`, instead keep the loading directly into the client instance TLS config
* Do not combine RootCA and ClientCA certs pool; keep them separate by nature
* 1 minute seems very frequent; certificate refresh only happens occasionally in a practical scenario. For now, let's keep 24 hours; then user can decide on it
* Keep pem refresher as part of client struct. Once it becomes part of the client, then we do not need lastChecked, log, debug, and certPool as part of the watcher

cdoucy:
Hi @jeevatkm thanks for the feedback!

> Make use of https://pkg.go.dev/time#Ticker to execute the function to reload based on a timer

yes, but that means that `ticker.Stop()` needs to be called at some point.
For this, `SetRootCertificateWatcher` may accept a `context.Context`, and `ticker.Stop()` will be called when the context is called.
What do you think?

> keep the loading directly into the client instance TLS config

Note sure to understand what you mean here.
To avoid using `OnBeforeRequest`, I can implement an `http.RoundTripper` that would get the refreshed Certificate and pass it to the `TLSClientConfig` of the underlying `http.Transporter`.
What do you think of this approach?


P.S. : I wish I could leverage `tls.Config.GetCertificate`, but it only works on the server side.

jeevatkm:
@cdoucy I'm not 100% sure, do we really need a context for this use case?

I may try to explain an approach based on what I learned from your PR and currently what is coming to my mind. This is not perfect; I am just dumping my thoughts.

```go
type CertWatcherOptions struct {
	// Default interval is 24 hours
	PoolInterval time.Duration
}

func (c *Client) SetRootCertificateWatcher(certPath string, options *CertWatcherOptions) *Client {
	c.SetRootCertificate(certPath)
	c.initRootCertWatcher(options)
 	return c
}

func (c *Client) SetClientRootCertificateWatcher(certPath string, options *CertWatcherOptions) *Client {
	c.SetClientRootCertificate(certPath)
	c.initClientRootCertWatcher(options)
 	return c
}

func (c *Client) initRootCertWatcher() {
	c.rootCAWatcher = &pemWatcher{
		// ...
	}

	// ...
}

func (c *Client) initClientRootCertWatcher() {
	c.clientRootCAWatcher = &pemWatcher{
		// ...
	}

	// ...	
}
```
So, that user could call `SetRootCertificateWatcher` method to set `N` number cert files, and those get added to the watch list
...
Let's assume there is a method `c.refreshRootCerts()` and `c.refreshClientRootCerts()` gets triggered by `time.Ticker` and reloads the certs into `c.tlsConfig().RootCAs` and `c.tlsConfig().ClientCAs`, respectively.

We can add a method to the Client Stop cert watcher ticker and add documentation for users to call that method from their shutdown hook scenarios.

What do you think?

jeevatkm:
@cdoucy, I thought I would check with you about the last comment.

cdoucy:
Hello @jeevatkm,

Thanks for your feedback! Sorry for the delayed response.
Your approach makes sense to me.
If a `Stop()` method can be introduced in the client, then we don't need a context.

I'll work on applying your suggestion.

jeevatkm:
@cdoucy, Can you please implement it against the main branch for Resty v3? I plan to introduce the method `Close` at the client level to tackle all the channels, tickers, etc. Clean-up could be done on that. 

Then the user would do something like this v3 onwards -
```go
client := resty.New()
defer client.Close()
...
```

cdoucy:
Hi @jeevatkm , sure!
I'll work on this over the weekend.

cdoucy:
Hi @jeevatkm !

Hope you are doing well.

I've made the following change to the PR:
- Rebased the branch off `main`
- Moved reloader code to client struct
- Refactored the code to use a `time.Ticker`
- Set the default pooling interval to 24 hours

client.go
```diff
@@ -208,6 +212,8 @@ type Client struct {
 	contentTypeDecoders      map[string]ContentTypeDecoder
 	contentDecompressorKeys  []string
 	contentDecompressors     map[string]ContentDecompressor
+	stopChan                 chan bool
+	certLock                 *sync.Mutex
```
jeevatkm, line 216: @cdoucy In Resty v3, RWMutex was added at the client level to protect all read/write operations on client fields. Please use that lock instead of creating a new one for the cert alone.

client.go
```diff
@@ -208,6 +212,8 @@ type Client struct {
 	contentTypeDecoders      map[string]ContentTypeDecoder
 	contentDecompressorKeys  []string
 	contentDecompressors     map[string]ContentDecompressor
+	stopChan                 chan bool
```
jeevatkm, line 215: @cdoucy, I wonder if we need an additional channel to stop the ticker. I will think about it.

jeevatkm (changes requested):
@cdoucy Thanks for updating PR. The implementation looks good.
I have added one comment and one thought; please have a look.

client.go
cdoucy: I removed it.

Since `handleCAs` is now goroutine safe, I don't even to acquire the lock in the cert watcher loop :)

client.go
cdoucy: The other approach I considered is to maintain a slice that would keep track of all `time.Ticker` instances, and the `Close()` method would iterate over this slice and stop all the tickers.

I judged that the channel approach is easier to implement and more in line with golang idiomatics.

Do you see any other approaches?

jeevatkm (approved):
@cdoucy Thanks for updating the PR.

jeevatkm:
@cdoucy It seems there is a test case failure. Can you please take care of it?

client.go
jeevatkm: I didn't get a chance to look at it deeper. We can keep the current approach. If needed, I can improve it afterward.

cdoucy:
> @cdoucy It seems there is a test case failure. Can you please take care of it?

Hi @jeevatkm I may have fixed it with 0959aa0b37ed4ff126d1c3d9ea330d3ba789a619

jeevatkm:
@cdoucy Thanks for fixing the test cases.  Can you please look at the missed lines in the test coverage?
https://github.com/go-resty/resty/pull/866/checks?check_run_id=31794960778

jeevatkm:
@cdoucy, I thought I would check with you. Is there any good news on the test case and its coverage?

cdoucy:
> @cdoucy, I thought I would check with you. Is there any good news on the test case and its coverage?

Hi @jeevatkm , just committed c6a6d2c7e563d4f80f0bf7d8f2409a8a1f8baf19 to improve coverage

jeevatkm (approved):
@cdoucy Thanks for the PR update.

PR #902, opened 2024-11-06
feat!: add fully composable request and response middleware

### LINKED ISSUE #345: RFC: create client let us choice not much more middleware.

### LINKED ISSUE #665: Multiple PreRequestHooks
I face the same problem as in https://github.com/go-resty/resty/issues/517#issue-1119261088 - I couldn't get not nil raw request.

The only way to get raw request is to use PreRequestHook. But we can have only one hook which will we overwritten if we try to attach the second one like this:
```
client := resty.New().
		SetPreRequestHook(preRequestHook1).
		SetPreRequestHook(preRequestHook2)
```

But having several hooks can be handy. I personally use it to:
1. Add custom auth which requires making hash of the request
2. Printing the final request as a curl (for debug purposes)
3. Can be used to remove headers which are set automatically like user-agent or content-type

That is why I wrote function which acts like this:

```
	client := resty.New().
		SetPreRequestHook(
			And(
				preRequestHook1,
				preRequestHook2,
			),
		)
```

So I want so propose to make it possible to have multiple PreRequestHook objects. That will allow to use resty this way:
 ```
client := resty.New().
		SetPreRequestHook(preRequestHook1).
		SetPreRequestHook(preRequestHook2)
```

### LINKED ISSUE #747: support OnBeforeResponse
```go
package main

import (
	"errors"
)

func main() {

	client := resty.New()
	
	client.OnBeforeResponse(func(client *resty.Client, response *resty.Response) error {

		if response.Size() > 1073741824 {

			return errors.New("file is too big")
		}

		return nil

	}).R().SetOutput("/path/123.zip").Get("http://www.ooxx.com/123.zip")

}

```

I want to determine the size of the content before parsing the Response, and throw an error if it exceeds my preset size.

### LINKED ISSUE #811: Client: Make middleware composable

jeevatkm:
- PreRequestHook removed
- Resty request and response middleware become exported
- Any order of middleware insertion is feasible

Closes #811
Closes #345
Closes #665
Closes #747

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,7 +517,7 @@ Resty provides middleware ability to manipulate for Request and Response. It is 
 client := resty.New()
 
 // Registering Request Middleware
-client.OnBeforeRequest(func(c *resty.Client, req *resty.Request) error {
+client.OnBeforeRequest(func(c *resty.Client, req *resty.Request) error { // TODO update docs
     // Now you have access to Client and current Request object
     // manipulate it as per your need
 
@@ -525,7 +525,7 @@ client.OnBeforeRequest(func(c *resty.Client, req *resty.Request) error {
   })
 
 // Registering Response Middleware
-client.OnAfterResponse(func(c *resty.Client, resp *resty.Response) error {
+client.OnAfterResponse(func(c *resty.Client, resp *resty.Response) error { // TODO update docs
     // Now you have access to Client and current Response object
     // manipulate it as per your need
 
--- a/client.go
+++ b/client.go
@@ -87,9 +87,6 @@ type (
 	// ResponseMiddleware type is for response middleware, called after a response has been received
 	ResponseMiddleware func(*Client, *Response) error
 
-	// PreRequestHook type is for the request hook, called right before the request is sent
-	PreRequestHook func(*Client, *http.Request) error
-
 	// DebugLogCallback type is for request and response debug log callback purpose.
 	// It gets called before Resty logs it
 	DebugLogCallback func(*DebugLog)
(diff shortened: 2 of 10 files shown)


arvenil:
Hello,

1. Any chance this could be back-ported to v2?
2. How far v3 is from stable release?

PR #934, opened 2025-01-01
feat(enhancement)!: seperate generate curl cmd from debug flow #928

### LINKED ISSUE #928: Why GenerateCurlCommand with debug
I require the GenerateCurlCommand function to log information, but I do not wish to include any debug-related output. Previously, I had written a function named restyResult2curl, but in the newer version, I discovered the GenerateCurlCommand. However, recently, GenerateCurlCommand underwent changes. Specifically, the EnableTrace() method is no longer available, and setting the debug level results in excessive output. 
😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠😠

jeevatkm:
fixes #928

### DIFF
--- a/client.go
+++ b/client.go
@@ -209,7 +209,8 @@ type Client struct {
 	proxyURL                 *url.URL
 	requestDebugLog          DebugLogCallback
 	responseDebugLog         DebugLogCallback
-	generateCurlOnDebug      bool
+	generateCurlCmd          bool
+	debugLogCurlCmd          bool
 	unescapeQueryParams      bool
 	loadBalancer             LoadBalancer
 	beforeRequest            []RequestMiddleware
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -1965,35 +1967,56 @@ func (c *Client) SetTrace(t bool) *Client {
...
 // NOTE: Use with care.
-//   - Potential to leak sensitive data from [Request] and [Response] in the debug log.
-//   - Beware of memory usage since the request body is reread.
+//   - Potential to leak sensitive data from [Request] and [Response] in the debug log
+//     when the debug log option is enabled.
+//   - Additional memory usage since the request body was reread.
+//   - curl body is not generated for [io.Reader] and multipart request flow.
+//
+// It can be overridden at the request level; see [Request.SetGenerateCurlCmd]
+func (c *Client) SetGenerateCurlCmd(b bool) *Client {
```
ccoVeille, line 2006: This one shouldn't be exported.

Otherwise, you had no reason to add Enable and Disable one,no?

Or maybe I'm missing something

jeevatkm: I kept two kinds of usage variation: Resty users choose one approach based on their need/preference. You may see this in nature methods for frequently used ones, e.g. Debug, Trace, curl, and retry default conditions.

PR #953, opened 2025-01-15
fix: retry not reset multipartField reader

### LINKED ISSUE #952: retry not reset multipartFields reader when RetryResetReaders is true
retry not reset multipartFields reader when RetryResetReaders is true

wxip:
fix #952


jeevatkm (changes requested):
@wxip Thanks for the PR. Can you please add test cases to cover PR changes?
See here - https://github.com/go-resty/resty/pull/953#issuecomment-2596399938

jeevatkm (approved):
@wxip Thanks for adding test cases.

PR #985, opened 2025-03-13
feat[sse]: make method and body configurable

vegetablechicken233:
Make HTTP Method configurable and add corresponding test functions.
It should be noted that to ensure backward compatibility, the "method" field does not perform empty validation, but is set to "GET" by default (to align with the old version).
 In addition, type of "body" field is temporarily set to "io.Reader" (it seems a bit too early to add support for "any"~

### DIFF
--- a/sse.go
+++ b/sse.go
@@ -368,8 +368,8 @@ func (es *EventSource) Get() error {
 		es.method = defaultHTTPMethod
 	}
 
-	if _, found := es.onEvent[defaultEventName]; !found {
-		return fmt.Errorf("resty:sse: OnMessage function is required")
+	if len(es.onEvent) == 0 {
+		return fmt.Errorf("resty:sse: At least one OnMessage/AddEventListener func is required")
 	}
 
 	// reset to begin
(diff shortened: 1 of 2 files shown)


jeevatkm (dismissed):
@vegetablechicken233 Thanks for the PR.

vegetablechicken233:
BTW I use "len(es.onEvent) == 0 " check instead of "if _, found := es.onEvent[defaultEventName]" check.

Sometimes the implementation of the sse server is not standard, and there is no event with the type of message.
In my opinion, users should be able to use it as long as they have added any eventListener.

@jeevatkm

jeevatkm:
> BTW I use "len(es.onEvent) == 0 " check instead of "if _, found := es.onEvent[defaultEventName]" check.
> 
> Sometimes the implementation of the sse server is not standard, and there is no event with the type of message. In my opinion, users should be able to use it as long as they have added any eventListener.
> 
> @jeevatkm

@vegetablechicken233 Can you please create a separate PR for this method check changes?

vegetablechicken233:
@jeevatkm  I created two individual PR for each feature:
https://github.com/go-resty/resty/pull/988
https://github.com/go-resty/resty/pull/989

Please take the time to review them if you have time. If everything is fine, please close this PR.

jeevatkm:
Closing this PR in favor of PR #988 and #989

PR #988, opened 2025-03-21
[feat] sse client support method and body configure

vegetablechicken233:
Make HTTP Method configurable and add corresponding test functions.
It should be noted that to ensure backward compatibility, the "method" field does not perform empty validation, but is set to "GET" by default (to align with the old version).
In addition, type of "body" field is temporarily set to "io.Reader" (it seems a bit too early to add support for "any"~

Separated from PR  https://github.com/go-resty/resty/pull/985

### DIFF
--- a/sse.go
+++ b/sse.go
@@ -152,7 +152,8 @@ func (es *EventSource) SetHeader(header, value string) *EventSource {
 
 // SetBody method sets body value to the [EventSource] instance
 //
-//	es.SetBody([]byte(`{"test":"put_data"}`),)
+// Example:
+// es.SetBody(bytes.NewReader([]byte(`{"test":"put_data"}`)))
 func (es *EventSource) SetBody(body io.Reader) *EventSource {
 	es.body = body
 	return es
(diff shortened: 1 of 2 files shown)


sse.go
```diff
@@ -139,6 +150,14 @@ func (es *EventSource) SetHeader(header, value string) *EventSource {
 	return es
 }
 
+// SetBody method sets body value to the [EventSource] instance
+//
+//	es.SetBody([]byte(`{"test":"put_data"}`),)
```
jeevatkm: @vegetablechicken233 This good example requires a correction based on method input type. Also, it has a syntax issue: `,` is present near the closing curve bracket.
```go
es.SetBody(strings.NewReader(`{"test":"put_data"}`))
```

jeevatkm (changes requested):
@vegetablechicken233 Thanks for creating a new PR. Just one minor comment on godoc. Once fixed, I will get it merged.

sse.go
vegetablechicken233: @jeevatkm  already fix it

jeevatkm (dismissed):
Thanks for the updates.

jeevatkm (approved):

PR #989, opened 2025-03-21
[refactor] use AddEventListener check instead of OnMessage check

vegetablechicken233:
Use "len(es.onEvent) == 0 " check instead of "if _, found := es.onEvent[defaultEventName]" check.

Sometimes the implementation of the sse server is not standard, and there is no event with the type of message.
In my opinion, users should be able to use it as long as they have added any eventListener.

Separated from PR https://github.com/go-resty/resty/pull/985

### DIFF
--- a/sse.go
+++ b/sse.go
@@ -343,8 +343,9 @@ func (es *EventSource) Get() error {
 	if isStringEmpty(es.url) {
 		return fmt.Errorf("resty:sse: event source URL is required")
 	}
-	if _, found := es.onEvent[defaultEventName]; !found {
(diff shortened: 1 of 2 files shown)


jeevatkm (commented):
@vegetablechicken233 Thanks for the PR.

jeevatkm:
@vegetablechicken233, there is a test case failure, please fix.
https://github.com/go-resty/resty/pull/989/checks

jeevatkm (approved):
Thanks for fixing the test case.

PR #990, opened 2025-03-21
fix: handle nil credentials before clone

gtwn:
can not clone parent client if not set credentials


jeevatkm:
@gtwn Thanks for the PR.

PR #997, opened 2025-04-07
Fix doc of SetRawPathParam and SetPathParam

icbd:
Fix doc about methods:
- SetRawPathParam
- SetPathParam


jeevatkm (approved):
Thanks for fixing godoc.

Issue #1001, opened 2025-04-15
decodeJSON will hang


Khachi-at:
- Test codes as follow: 

```go
func TestXxx(t *testing.T) {
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("null"))
	}))

	client := resty.New().SetRetryCount(DefaultRetryTimes).EnableGenerateCurlCmd()

	x := map[string]string{}

	resp, err := client.R().SetBody("{}").
		SetHeader("Content-Type", "application/json; charset=utf-8").
		SetForceResponseContentType("application/json").
		SetAllowMethodGetPayload(true).
		SetResponseBodyUnlimitedReads(true).
		SetResult(x).
		Get(server.URL + "/test")

	fmt.Println(err)
	fmt.Println(resp.String())
}
```

- The decodeJson function will loop infinitely

<img width="1288" alt="Image" src="https://github.com/user-attachments/assets/64123936-4d41-46c4-a1e1-da69ab0dbdf7" />

Khachi-at:
I think the bug is at there. After reading the body, no write back to r.Body.

<img width="821" alt="Image" src="https://github.com/user-attachments/assets/5aa8a973-826b-47ca-88a4-d224600eb0fe" />

Khachi-at:
The following code can make this bug clearer：

```golang
package resty

import (
	"io"
	"strings"
	"testing"
)

func TestDecodeJSON(t *testing.T) {
	r := &Response{
		Body: io.NopCloser(strings.NewReader("null")),
	}
	r.wrapCopyReadCloser()
	err := r.readAll()

	assertNil(t, err)

	var result string
	err = decodeJSON(r.Body, &result)
	assertNil(t, err)
	assertEqual(t, "null", result)
}

```

hu-xiaokang:
I think the problem is here: `r.r.Seek(0, io.SeekStart)`. Why do we need to do this?

```
type nopReadCloser struct {
	r *bytes.Reader
}

func (r *nopReadCloser) Read(p []byte) (int, error) {
	n, err := r.r.Read(p)
	if err == io.EOF {
		r.r.Seek(0, io.SeekStart)
	}
	return n, err
}
```

jeevatkm:
Fixed by PR #1002

PR #1002, opened 2025-04-16
fix: Fixed functions of decodeJson do not return.

Khachi-at:
Fixed the [issuce](https://github.com/go-resty/resty/issues/1001)

### DIFF
--- a/response.go
+++ b/response.go
@@ -237,7 +237,7 @@ func (r *Response) readAll() (err error) {
 	} else {
 		r.bodyBytes, err = ioReadAll(r.Body)
 		closeq(r.Body)
-		r.Body = &nopReadCloser{r: bytes.NewReader(r.bodyBytes)}
+		r.Body = &nopReadCloser{r: bytes.NewReader(r.bodyBytes), resetOnEOF: true}
 	}
 	if err == io.ErrUnexpectedEOF {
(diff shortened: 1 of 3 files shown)


response.go
```diff
@@ -265,7 +265,7 @@ func (r *Response) wrapCopyReadCloser() {
 		f: func(b *bytes.Buffer) {
 			r.bodyBytes = append([]byte{}, b.Bytes()...)
 			closeq(r.Body)
-			r.Body = &nopReadCloser{r: bytes.NewReader(r.bodyBytes)}
+			r.Body = io.NopCloser(bytes.NewReader(r.bodyBytes))
```
jeevatkm: @Khachi-at `nopReadCloser` has a functionality. If any issues, we have to fix that instead of replacing it with `io.NopCloser`.
If you're interested in finding a fix, kindly update the PR. Otherwise, let me know. I will have a look at #1001.

jeevatkm (changes requested):
@Khachi-at Thanks for the PR. Please see my comment.

response.go
Khachi-at: @jeevatkm Hello, the PR has been updated.

jeevatkm: Thanks @Khachi-at for addressing.

jeevatkm (approved):

Issue #996, opened 2025-03-30
Content-Length header missing for empty POST requests


N1cOs:
Hello! Recently we faced with a problem, that resty Client doesn't set `Content-Length` header for an empty POST request.

Version - 2.16.5. 

Example:
```go
package main

import (
	"fmt"

	"github.com/go-resty/resty/v2"
)

func main() {
	client := resty.New()

	client.OnBeforeRequest(func(client *resty.Client, request *resty.Request) error {
		request.
			SetBody(nil).
			SetContentLength(true)

		return nil
	})

	client = client.SetBaseURL("http://localhost:8080")

	resp, err := client.R().Execute("POST", "/test")
	if err != nil {
		panic(err)
	}

	fmt.Println(resp)
}
```
Then I captured request with Wireshark and `Content-Length` header is missing:

<img width="971" alt="Image" src="https://github.com/user-attachments/assets/5d6ebd4f-fa9d-47e3-bc58-7aa09ce6cc8e" />

N1cOs:
I did a little more research. It happens not only for POST requests but also for PUT, PATCH. 

From version v2.16.2 there is a workaround to use `http.NoBody` explicitly.

Libero-Dev:
@N1cOs I have created a [PR](https://github.com/go-resty/resty/pull/1003) for this. Ultimately I'm not so sure if this is something that should or shouldn't be merged in and supported as the omission of the Content-Length header seems to be something of an undocumented (that or documentation on this is very difficult to find) side effect in the standard http library when the resty request is built with http.NewRequest.

N1cOs:
@Libero-Dev thanks for the fix. IMO it should be fix. Standard http library has the logic for handling a nil body. For instance, this code snippet writes Content-Length header in the request:
```go
package main

import "net/http"

func main() {
	_, err := http.Post("http://localhost:8080/test", "application/json", nil)
	if err != nil {
		panic(err)
	}
}
```
Library sends the request:
```txt
POST /test HTTP/1.1
Host: localhost:8080
User-Agent: Go-http-client/1.1
Content-Length: 0
Content-Type: application/json
Accept-Encoding: gzip
```

jeevatkm:
This is fixed by PR #1003

PR #1003, opened 2025-04-19
Fix/missing content length header on empty post req

Libero-Dev:
This change aims to resolve issue #996 by changing requests that contain nil body to http.NoBody when they have also specified to "SetContentLength" as true. 

Validity as to whether to merge this or not depends on multiple factors.

1. The root issue as to why Content-Length was ommitted when requests were created with a nil body ultimately originates from within the go http standard library. The library explicitly states that Content-Length will be omitted when body is nil unless Transfer-Encoding header is explicitly set to "identity". Where exactly does this happen in the http library? I have no idea, despite trying to find it. However I have found similar issues that reference to this: https://github.com/golang/go/issues/40978.

Setting the bodyBuf value to http.NoBody signals to the http library when creating the request that the body is no longer nil and a Content-Length header component can be created for the request.

3. Considering this is a feature, albeit a poorly documented one in the standard http library it does seem to be behaving as intended?

4. In resty setting SetContentLength(true) and r.Body = nil, could be considered as the user intentionally overriding this requirement to force the Content-Length header on whilst passing in a nil body.

Ultimately, either the intention of the http libraries functionality should be obeyed or we should consider this unique resty request setting combination as an intentional force override to achieve a valid Content-Length header with a nil body.

Additional Notes:

I have had to update middleware_test.go:Test_parseRequestBody() for the reason being that now that nil body + SetContentLength(true) is supported in this PR change this now results in the expectedBodyBuf value to change since in this scenario the bodyBuf changes to http.NoBody.

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -188,12 +188,12 @@ func parseRequestHeader(c *Client, r *Request) error {
 }
 
 func parseRequestBody(c *Client, r *Request) error {
-	// Go http library omits Content-Length if body is nil; use http.NoBody to force it when SetContentLength is true
-	if r.Body == nil && (c.setContentLength || r.setContentLength) {
-		r.Body = http.NoBody
-	}
-
 	if isPayloadSupported(r.Method, c.AllowGetMethodPayload) {
+		// Go http library omits Content-Length if body is nil; use http.NoBody to force it when SetContentLength is true
+		if (r.Body == nil && r.bodyBuf == nil) && (c.setContentLength || r.setContentLength) {
+			r.Body = http.NoBody
+		}
+
 		switch {
 		case r.isMultiPart: // Handling Multipart
 			if err := handleMultipart(c, r); err != nil {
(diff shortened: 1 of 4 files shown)


jeevatkm (changes requested):
@Libero-Dev I have read the issue #996 and this PR description along with the code changes. Thanks for the details.

Can you please have a look at these lines and apply the fix accordingly instead of adding new repeated code lines to fix?
https://github.com/go-resty/resty/blob/1e19d6bc6a102df59e1997a09850d0c0532e7f9a/middleware.go#L208-L215

Libero-Dev:
@jeevatkm there is reliance on the handleRequestBody() function being triggered before proceeding with the rest of the logic. Placing the conditional check where you have recommended will break tests and continue to exhibit missing content-length header for recipients of the request. I have compromised with a cleaner alternative that attempts to keep both requirements.

jeevatkm (approved):
@Libero-Dev Thanks for the updates.

PR #1010, opened 2025-04-24
feat(enhancement): change retry duration from seconds to milliseconds in SSE (#1010)

liuzengh:
Change the retry duration from seconds to milliseconds for proper handling of client-specified retry intervals. This ensures compatibility with standard SSE behavior and prevents potential timeout issues.
- https://html.spec.whatwg.org/multipage/server-sent-events.html#concept-event-stream-reconnection-time
- https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#retry


liuzengh:
@jeevatkm please take a look

jeevatkm (approved):
@liuzengh Thanks for the PR.

PR #1035, opened 2025-06-14
feat: add jaeger tracer

HuckOps:


jeevatkm:
@HuckOps Thanks for the PR. I'm sorry for the delayed response, I was occupied with my personal stuff.

I will not add any external library into Resty directly; instead, I do an interface approach so that Resty users could integrate any library that suits its requirements.

PR #1056, opened 2025-09-24
Use crypto/sha256 instead of crypto/md5 for FIPS compliance.

BillBuilt:
Use crypto/sha256 instead of crypto/md5 for FIPS compliance.

Update deps

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,5 @@
 module resty.dev/v3
 
-go 1.24.0
+go 1.21
 
-toolchain go1.24.6
-
-require golang.org/x/net v0.44.0
+require golang.org/x/net v0.33.0
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module resty.dev/v3
 
-go 1.21
+go 1.24.0
 
-require golang.org/x/net v0.33.0
+toolchain go1.24.6
+
+require golang.org/x/net v0.44.0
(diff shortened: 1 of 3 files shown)


smithcoin:
Not only for fips, many linters now suggest avoiding md5 for any reason.

moorereason:
The comment in line 374 should be updated to not reference md5.

What are the implications of this change?  From what I can see, the machine ID is only used here:
https://github.com/go-resty/resty/blob/66256efbe00678090075a7862bf4113c8e183c8d/util.go#L375
Which is only called from here:
https://github.com/go-resty/resty/blob/66256efbe00678090075a7862bf4113c8e183c8d/request.go#L1385

BillBuilt:
When running go code using `GOFIPS140=v1.0.0 GODEBUG=fips140=only`, if any package is not compliant (MD5), it will fail to build regardless of how it is used. 

https://go.dev/doc/security/fips140

I will update the comment. Thanks!

jeevatkm (changes requested):
@BillBuilt Thanks for the PR. I'm sorry for the delayed response, I was occupied with my personal stuff.

Can you please revert the `go.{mod, sum}` file changes?

BillBuilt:
> @BillBuilt Thanks for the PR. I'm sorry for the delayed response, I was occupied with my personal stuff.
> 
> Can you please revert the `go.{mod, sum}` file changes?

No worries. I reverted the version changes. I did get this warning so maybe consider updating any dependencies when you get a chance. 

```Warning:(5, 9)  Dependency go:golang.org/x/net:v0.33.0 is vulnerable CVE-2025-22870 4.4 ```


Thanks again!

jeevatkm (approved):
@BillBuilt Thanks for updating the PR.

jeevatkm:
@BillBuilt Sure, I usually update the go.mod during the release.

PR #1057, opened 2025-09-29
FIPS compliance.

BillBuilt:
Use crypto/sha256 instead of crypto/md5 for FIPS compliance.

Update deps


BillBuilt:
apologies - looks like this PR was unnecessary - refer to https://github.com/go-resty/resty/pull/1056

PR #1038, opened 2025-06-17
fix: negative trace substraction when using SetTimeout

nicolasbeauvais:
In certain cases when using [client](https://pkg.go.dev/resty.dev/v3#Client.SetTimeout) or [request](https://pkg.go.dev/resty.dev/v3#Request.SetTimeout) SetTimeout it is possible that some trace value like `dnsDone`, `tlsHandshakeDone` or `gotConn` are zero, resulting in inaccurate results:

Screenshot From 2025-06-18 01-19-03
Screenshot From 2025-06-18 01-18-54
Screenshot From 2025-06-18 01-18-12

This PR add additional checks in the `TraceInfo` method to avoid negative values.

I'm not sure how to improve the tests for this, as aborting requests with SetTimeout will not always accurately stop the execution at the exact same point. I'm open to ideas.

PR #1064, opened 2025-11-19
Add HTTPS request with trace test for coverage

nicolasbeauvais:
This test increase coverage of the changes introduced in https://github.com/go-resty/resty/pull/1038


jeevatkm (approved):
@nicolasbeauvais Thanks for the help to bring back the coverage.

PR #1098, opened 2026-01-01
Add support for `SetXXXAny` methods for  headers, query parameters, and path parameters.

### LINKED ISSUE #1027: Add `SetQueryParamInt`/`Bool` methods to directly serialize bool and int for `resty.Request`
Right now the SetQueryParam method only accepts `string`, which forces developers to write conversion code to reformat `bool`s and `int`s to string. `bool` is a very common query parameter type used for filtering results from an endpoint, while `int` is used all the time for things like pagination.

To make resty easier to use with these common go types, I would like to add two methods to the client:
`SetQueryParamInt(param string, value int)`
and
`SetQueryParamBool(param string, value bool)`.

These can call `strconv.Itoa` and `strconv.FormatBool` in their implementation to generate the required strings for the query param url.

EDIT: If we can add a similar method for Path Params like `SetPathParamInt`, that would be great as well

pscheid92:
Add support for SetXXXAny methods to simplify handling various data types in headers, query parameters, and path parameters.

Closes #1027

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"strings"
 	"testing"
+	"time"
 )
 
 func Benchmark_parseRequestURL_PathParams(b *testing.B) {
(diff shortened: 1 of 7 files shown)


jeevatkm (approved):
@pscheid92 Thanks for the PR. 

Can you please add documentation? 
Resty v3 documentation is here: https://github.com/go-resty/docs

pscheid92:
Thank you @jeevatkm! And sure, the documentation is already available here:
https://github.com/go-resty/docs/pull/18

PR #1123, opened 2026-03-13
Add SetLabel() to be discoverable to metrics

xxtanisxx:
Adding `request.SetLabel()`. The idea is to read that label in `responseMiddleware` and emit Prometheus histogram metrics with the request duration and label.

Many of the URLs include dynamic path segments, so using the raw URL would introduce high-cardinality metrics which increase cost.

https://github.com/go-resty/resty/discussions/1122 for the discussion

### DIFF
--- a/request.go
+++ b/request.go
@@ -73,6 +73,7 @@ type Request struct {
 	RetryDelayStrategy           RetryDelayStrategyFunc
 	IsRetryDefaultConditions     bool
 	IsRetryAllowNonIdempotent    bool
+	Label                        string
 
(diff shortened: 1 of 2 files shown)


jeevatkm (dismissed):
@xxtanisxx Thanks for the PR.

jeevatkm:
@xxtanisxx It seems `go fmt` fails. Can you check the CI build and fix it?

xxtanisxx:
@jeevatkm Thanks, just ran `go fmt`. Hopefully this help pass the test now.

jeevatkm (approved):

Issue #334, opened 2020-05-04
Retry doesn't work with FileReader and MultipartFields


phsym:
Hi,
First, thank you for this excellent library.

### Description
I'm trying to send data using `FileReader `and `MultipartFields`, but the server I'm sending those data to is a bit crappy, and sometimes drops the connections. Unfortunately I have no control on this server.
I configured the resty client to use the retry mechanisms, which works well for almost all my requests, except those with `FileReader `and `MultipartFields`.
Each retry attempt is having an `EOF `error because the `io.Reader` has been consumed already.

I believe this issue will happen also with `Request.Body` implementing `io.Reader`, but I have not checked.

### Possible fix
A possible fix would be to check if the readers implement `io.Seeker`, and seek back to the beginning. But maybe we should consider the case where a reader is not read from the beginning, thus seeking to offset 0 would be problematic. What about keeping the number of bytes read, and seek back to the original offset ?

Other alternative, probably more efficient, would be to keep `Request.bodyBuf` between attempts, not trying to create it each time, then releasing this buffer only on success.

### Workaround
In the meantime, as a workaround, I'm writing my data to a temporary file, and I use the `SetFile `method.

jeevatkm:
@phsym I believe your understanding is correct. Resty keeps complete request in the memory before it sends to the server, this approach has downside on large request body scenarios, that's why `io.Reader` introduced and resty submits the body handling to underlying Go HTTP client if its an `io.Reader`.

I'm not sure, keeping `Request.bodyBuf` around all retry. what will be the side effect? I have to try out and see.

I will have to think about, how to further improve the request body handling for various use cases supported by resty. Right now I don't have an answer. I'm sorry!

jeevatkm:
Just referencing an issue #309

LBeernaertProton:
We recently ran into this issue, I have made a fix in a fork which suits our current needs. Let me know if you would be interested in a pull request based of those changes. 

Code is available [here](https://github.com/LBeernaertProton/resty/tree/multipart-streams).

jeevatkm:
This has been taken care of as part of #879, and Resty automatically seeks if the provided reader has an `io.ReadSeeker` interface.

PR #1127, opened 2026-03-16
feat: add ReaderFactory for non-seekable multipart reader retry

### LINKED ISSUE #770: The request with SetFileReader  retry time great than 1 will has issues
Below is the SetFileReader function code
```
func (r *Request) SetFileReader(param, fileName string, reader io.Reader) *Request {
	r.isMultiPart = true
	r.multipartFiles = append(r.multipartFiles, &File{
		Name:      fileName,
		ParamName: param,
		Reader:    reader,
	})
	return r
}
```
When we first use the reader to read content, the reader current index will move to the end. When we retry it again, the current index is in the end, we will lost the content from the start to the end. This means the second time we read the content is empty.

mcanevet:
## Summary

- Add `ReaderFactory` interface for multipart fields that creates a fresh `io.Reader` per retry attempt, solving the non-seekable reader retry gap (#770, #334)
- Add `ErrReaderNotSeekable` sentinel error so non-seekable readers without a factory fail loudly instead of silently sending corrupt bodies on retry

## Motivation

PR #549 added `io.ReadSeeker` support for resetting multipart readers on retry. However, non-seekable readers (cipher streams, network pipes, compression wrappers) had no solution — `resetReader()` silently returned `nil`, causing retried requests to send empty or truncated bodies.

## New APIs

- `ReaderFactory` interface: single `NewReader() io.Reader` method
- `LenReaderFactory` interface: optional `Len() int64` for content length
- `BytesReaderFactory`, `StringReaderFactory`: built-in implementations (both implement `LenReaderFactory`)
- `NewMultipartFieldFromFactory(name, fileName, contentType, factory)`: constructor
- `MultipartField.Factory` field: set directly or via constructor
- `ErrReaderNotSeekable`: returned when a non-seekable reader has no factory (**breaking change** — previously returned `nil`)

## Breaking change

`resetReader()` now returns `ErrReaderNotSeekable` instead of `nil` for readers that don't implement `io.ReadSeeker` and have no `Factory` set. Code that passes `bytes.Buffer` or similar non-seekable readers with retries will now get an explicit error. The fix is to either:
1. Use `io.ReadSeeker` (e.g. `bytes.Reader` instead of `bytes.Buffer`), or
2. Use `NewMultipartFieldFromFactory` with a `ReaderFactory`

This is intentionally split into two commits for reviewability.

## Test plan

- [x] Unit tests for `BytesReaderFactory`, `StringReaderFactory`, `LenReaderFactory`
- [x] Unit test for `resetReader()` with factory (closes old reader, creates fresh one)
- [x] Unit test for custom `ReaderFactory` implementation
- [x] Integration test: retry with factory produces non-empty body on each attempt
- [x] Integration test: middleware called for each attempt with factory
- [x] Integration test: `ErrReaderNotSeekable` propagates through retry loop to caller
- [x] Full existing test suite passes

Fixes #770
Related: #334, #549

### DIFF
--- a/multipart.go
+++ b/multipart.go
@@ -7,6 +7,7 @@ package resty
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -17,6 +18,15 @@ import (
 )
 
 var quoteEscaper = strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
+
+// ErrReaderNotSeekable is returned when a multipart field reader
+// does not implement [io.ReadSeeker] and no [ReaderFactory] is provided.
+// This prevents silent data corruption during retry attempts where a
+// consumed [io.Reader] would produce an empty or truncated body.
+var ErrReaderNotSeekable = errors.New(
+	"resty: multipart reader is not seekable and no factory provided; " +
+		"use resty.NewMultipartFieldFromFactory for retry support with non-seekable readers",
+)
 
 // ReaderFactory is an interface for creating fresh [io.Reader] instances
 // per retry attempt. This is useful for non-seekable readers (cipher streams,
@@ -124,7 +134,7 @@ func (mf *MultipartField) resetReader() error {
 		return err
 	}
 
-	return nil
+	return ErrReaderNotSeekable
 }
 
 func (mf *MultipartField) isValues() bool {
--- a/multipart.go
+++ b/multipart.go
@@ -6,6 +6,7 @@
 package resty
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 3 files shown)


jeevatkm (commented):
@mcanevet, thanks for submitting the PR.

I’ve reviewed the PR changes.

The current Resty design works perfectly as long as the user implements the [io.ReadSeeker](https://pkg.go.dev/io#ReadSeeker) at their end. Therefore, introducing another new interface isn’t necessary. I regret to inform you that I can’t accept the current PR.

I prefer to keep the Resty side very minimal and leave the rest to the user.

However, I appreciate your suggestion, “fail loudly instead of silently.” I agree with you. In the upcoming major v3 release, we can consider implementing break changes.

Would you be interested in updating this PR to “introduce new error and propagate the error, instead of nil”? Please let me know.

PR #1133, opened 2026-03-22
fix: return error for non-seekable readers on request retry

mcanevet:
## Summary

- Previously `resetReader()` silently returned `nil` when a multipart field reader did not implement `io.ReadSeeker`. This caused retried requests to send empty or corrupt bodies without any error signal (#770).
- Now `resetReader()` returns `ErrReaderNotSeekable`, which propagates through the retry loop and surfaces to the caller.

## Breaking Change

This is a breaking change for code that passes non-seekable readers (e.g. `bytes.Buffer`) and relies on retries. Such code should either:
- Switch to `io.ReadSeeker` (e.g. `bytes.Reader`)
- Wait for a future release with `ReaderFactory` support

As suggested by @jeevatkm in https://github.com/go-resty/resty/pull/1127#issuecomment.

### DIFF
--- a/multipart.go
+++ b/multipart.go
@@ -6,6 +6,7 @@
 package resty
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -16,6 +17,8 @@ import (
 )
 
 var quoteEscaper = strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
+
+var ErrReaderNotSeekable = errors.New("resty: multipart reader is not seekable on request retry")
 
 func escapeQuotes(s string) string {
 	return quoteEscaper.Replace(s)
@@ -75,7 +78,7 @@ func (mf *MultipartField) resetReader() error {
 		_, err := rs.Seek(0, io.SeekStart)
 		return err
 	}
-	return nil
+	return ErrReaderNotSeekable
 }
 
 func (mf *MultipartField) isValues() bool {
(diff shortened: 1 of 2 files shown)


multipart.go
```diff
@@ -17,53 +18,60 @@ import (
 
 var quoteEscaper = strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
 
+var ErrReaderNotSeekable = errors.New(
```
jeevatkm: @mcanevet Can you please update the the error message to ?
```go
var ErrReaderNotSeekable = errors.New("resty: multipart reader is not seekable on request retry")
```

jeevatkm (changes requested):
@mcanevet Thanks for creating a PR. Please check the comment.

jeevatkm:
@mcanevet checking on requested changes.

mcanevet:
I'll try to find some time to do it this weekend. I was offline this week

mcanevet:
@jeevatkm done

jeevatkm:
@mcanevet Thanks for updating PR. It seems PR has test case failures.
```
=== Failed
=== FAIL: . TestRetryNonSeekableReaderWithoutFactoryReturnsError (0.00s)
    multipart_test.go:752: Expected [resty: multipart reader is not seekable on request retry], got [<nil>]. Message: 

```

jeevatkm (approved):
Thanks for the update @mcanevet.

PR #1156, opened 2026-04-24
Return ErrReaderNotSeekable instead of sending empty bodies on retry

pior:
If we `SetBody()` is called with a non-retriable reader, retries will be sent with empty body.

A similar bug with the multipart reader was fixed in https://github.com/go-resty/resty/pull/1133.
The original discussion: https://github.com/go-resty/resty/pull/1127.

This PR follows the same logic: return `ErrReaderNotSeekable` instead of silently sending an empty body.

Breaking: the actual text of `ErrReaderNotSeekable` is changed to make the error generic, `multipart` is removed.

### DIFF
--- a/client.go
+++ b/client.go
@@ -60,6 +60,11 @@ var (
 	// ErrUnsupportedRequestBodyKind is returned when the request body is of an unsupported kind.
 	ErrUnsupportedRequestBodyKind = errors.New("resty: unsupported request body kind")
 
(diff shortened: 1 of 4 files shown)


jeevatkm (changes requested):
@pior Thanks for the PR, can you please check patch coverage failure?

jeevatkm:
@pior ping, any updates?

jeevatkm (approved):
@pior Thanks for the PR.

PR #1158, opened 2026-04-30
Fix flaky test TestClient_SetRootCertificateWatcher

pior:
I noticed a flaky test: https://github.com/go-resty/resty/actions/runs/25158074604
And I was able to reproduce it locally with `go test -race -run TestClient_SetRootCertificateWatcher -count=20 -v`

Changes:
- wait for the certificate watcher polling interval after rotating or restoring certificates
- avoid racing the next request against the asynchronous watcher reload


jeevatkm (approved):
Thanks for improving test case.

Issue #1016, opened 2025-05-09
The total time in trace info may become MAX_INT64 in some invalid requests


git-hulk:
Currently, the request total is calculated by `EndTime` - `DNS Start Time` | `Get Conn Time`:

https://github.com/go-resty/resty/blob/036c5894765c5b60f2c6992b8a1accae124019f7/request.go#L1256-L1260

And in [Go's transport](https://github.com/golang/go/blob/49d6777d87a0abb3eda032da95eff024156835f7/src/net/http/transport.go#L685), it won't call the `GetConn` if the request is invalid(for example: an invalid URL format), so both of `DNS Start Time` | `Get Conn Time` are zero. So it will get MAX_INT64 if we use a valid time to sub the zero time.

PR #1161, opened 2026-05-17
fix: include retry wait in TraceInfo TotalTime

### LINKED ISSUE #1142: Resp.Request.TraceInfo().TotalTime not include time spend to retry
We had a problem understanding why our endpoints were taking 30 seconds to complete.
After read logs, we couldn't figure out why we had two database selects and a request to an external endpoint (via RESTY). 
It took 30 seconds to execute.
In the logs, we see that request to the external service:
trace_info.conn_time: 204ms,
trace_info.dns_lookup: 21ms
trace_info.total_time 363ms

We started building conspiracy theories about Postgres, CPU throttling, etc...

Then, with great difficulty, we managed to find the APM span for this rare request. 
We discovered that out of the 30 seconds of this request, all 30 were taken by RESY. Im investigate my resty client settings config, find timeout is set to 30 seconds  and retry.
Now the question is, why didn't this time show up in the total time?
My understanding is that total time is the time it took Resty to execute my request from the moment I called .Get(someURI) until RESTY return response. Or am I misunderstanding the meaning of total time?

wyf027:
## Summary

`TraceInfo().TotalTime` and `Response.Duration()` now cover the full request lifecycle including retry backoff, not only the final HTTP attempt.

## Changes

- Set `Request.StartTime` only on the first attempt (before `http.Client.Do`)
- Compute `TraceInfo.TotalTime` from `StartTime` when set; keep trace-timestamp fallback for invalid requests (#1016)
- Add `TestTraceInfoTotalTimeIncludesRetryWait`

Fixes #1142

### DIFF
--- a/client.go
+++ b/client.go
@@ -2471,7 +2471,9 @@ func (c *Client) execute(req *Request) (*Response, error) {
 
 	prepareRequestDebugInfo(c, req)
 
-	req.StartTime = time.Now()
+	if req.StartTime.IsZero() {
+		req.StartTime = time.Now()
+	}
 	resp, err := c.Client().Do(req.withTimeout())
 	// Cancel multipart context for io.Copy to stop reading/writing further
 	if req.isMultiPart && req.multipartCancelFunc != nil {
--- a/request.go
+++ b/request.go
@@ -1344,14 +1344,15 @@ func (r *Request) TraceInfo() TraceInfo {
 		ti.ServerTime = ct.gotFirstResponseByte.Sub(ct.gotConn)
 	}
 
-	// Calculate the total time accordingly when connection is reused,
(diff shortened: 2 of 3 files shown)


jeevatkm (commented):
@leno23 Thanks for trying this. I’ve clarified the question about the total time in this [response](https://github.com/go-resty/resty/issues/1142#issuecomment-4174693776).

The term “total time” in the trace information refers to DNS lookup, TCP connection time, server time, and other relevant durations. This is because trace information provides time information within a request at various stages for the request. That’s why the variable named “TotalTime” is included in the trace information object.

I believe this PR is not a suitable implementation. It will cause a change in behavior and potentially lead to unexpected results.

PR #1162, opened 2026-05-17
feat(enhancement): honor explicit zeros in min and max retry wait time

### LINKED ISSUE #992: Unintuitive handling of 0 min and max for jitter backoff
Currently 0 min and max wait times for jittered backoff automatically fall back to the defaults, see here: https://github.com/go-resty/resty/blob/v3/retry.go#L82-L87 I think this is quite unintuitive. 

Maybe one can argue that jitter with 0 backoff between doesn't make sense, but it would be a convenient way to get a constant delay without needing to set a custom stragegy, i.e. by setting wait time and max wait time both to 0.

wyf027:
## Summary

Fixes #992

When both `SetRetryWaitTime(0)` and `SetRetryMaxWaitTime(0)` are set, jitter backoff now uses a constant zero delay instead of replacing zeros with library defaults.

Unset minimum wait (`0` with a positive max) still maps to `defaultWaitTime` as before.

## Test plan

- [x] `go test -run 'TestRetryCoverage|TestClientRetry'`
- [x] Added `explicit zero min and max yields zero delay` subtest

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -125,12 +125,9 @@ func isDoNotRetryError(err error) bool {
 }
 
 func newBackoffWithJitter(min, max time.Duration) *backoffWithJitter {
-	// Explicit zero min and max yields constant zero delay between retries (#992).
-	if min == 0 && max == 0 {
+	if min == 0 && max == 0 { // GH#992
 		return &backoffWithJitter{
 			rnd: rand.New(rand.NewPCG(rand.Uint64(), rand.Uint64())),
-			min: 0,
-			max: 0,
 		}
 	}
 	if min <= 0 {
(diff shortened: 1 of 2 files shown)


jeevatkm (changes requested):
@leno23 Thanks for the PR. Please fix the failing test case and have 100% patch coverage.

jeevatkm:
@leno23 Any updates on fixing the failing test case and having 100% patch coverage?

jeevatkm:
@leno23 I didn't get a response. So I will update the PR and move forward.

jeevatkm (approved):
I have made changes to the PR.

PR #1163, opened 2026-05-26
fix: reuse encoded request body on retry attempts

### LINKED ISSUE #1160: post方式下，body会多次深copy
在post方式下， 如果body不是reader的话，会产生多次数据copy

wyf027:
## Summary

Fixes #1160

Non-`io.Reader` request bodies were re-encoded on every retry because the shared `bodyBuf` was drained by the HTTP transport after the first attempt.

Cache the encoded payload after the first encode and refill `bodyBuf` on later attempts.

## Test plan

- [x] Added `TestRetryReusesEncodedRequestBody`
- [x] `go test -run 'TestRetry|parseRequestBody'`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -460,13 +460,12 @@ func handleRequestBody(c *Client, r *Request) error {
 		r.Header.Set(hdrContentTypeKey, contentType)
 	}
 
-	r.bodyBuf = acquireBuffer()
-
-	switch body := r.Body.(type) {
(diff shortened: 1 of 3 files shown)


jeevatkm (commented):
@leno23 Thanks for attempting the PR. It is not a bug, it’s by design in the Resty, so that user could modify data between retry request data on hook functions.

PR #1164, opened 2026-05-26
fix: preserve trailing slash in SetBaseURL

### LINKED ISSUE #119: Trailing slash removed
After commit ad03978733316c, if a request is made with a trailing slash, it is removed in function `composeRequestURL ` (from util.go).
This is not expected, since there are many APIs that require trailing slashes to the endpoints.

### LINKED ISSUE #1061: v3: Regression of #119
Looks like at least when using `<resty.Client>.SetBaseURL()`, the trailing slash is removed.

This was previously addressed in v1.2 as reported in #119 and it seems to have reared its head again.

wyf027:
## Summary

Fixes #1061
Fixes #119

`SetBaseURL` used `strings.TrimRight` to strip trailing slashes, so `BaseURL()` no longer matched the configured value. This regressed the v1.2 behavior for callers that rely on a trailing slash when composing URLs.

## Test plan

- [x] Added `TestSetBaseURLPreservesTrailingSlash`
- [x] `go test -run 'TestSetBaseURLPreservesTrailingSlash|TestClientClone'`
- [x] `go test -run 'Test_parseRequestURL/using_BaseURL'`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/client.go
+++ b/client.go
@@ -299,7 +299,7 @@ func (c *Client) BaseURL() string {
 func (c *Client) SetBaseURL(url string) *Client {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	c.baseURL = strings.TrimRight(url, "/")
+	c.baseURL = url
 	return c
 }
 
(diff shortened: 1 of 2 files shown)


jeevatkm (commented):
@leno23 Thanks for attempting PR. Appreciate it.

It is not a bug, please see the details here - https://github.com/go-resty/resty/issues/1061#issuecomment-3650034579

PR #1165, opened 2026-05-26
fix: separate URL from -F/-d in buildCurlCmd output

c-tonneslan:
buildCurlCmd has two body branches that forget the trailing space that the normal `-d`-with-body branch adds, so the URL ends up glued to the previous token:

    curl -X POST -H 'Content-Type: multipart/form-data' -F '<fields omitted, see original request>'http://example.com/upload
    curl -X POST ... -d ''http://unexecuted-request

curl reads those as part of the `-F` value or `-d` argument and the command isn't usable. The other branch (normal body) already appends a space, so it's just an inconsistency in two spots.

Also tightened the multipart test and `TestCurlRequestGetBodyError` to actually check there's a space before the URL. The previous `strings.Contains(curlCmd, tt.url)` and `strings.Contains(curlCmdUnexecuted, "-d ''")` matched fine even with the URL glued on, which is how this slipped through.

### DIFF
--- a/curl.go
+++ b/curl.go
@@ -46,7 +46,7 @@ func buildCurlCmd(req *Request) string {
 	contentType := req.RawRequest.Header.Get(hdrContentTypeKey)
 	if strings.HasPrefix(contentType, "multipart/form-data") {
 		// Multipart: show placeholder
-		curl += "-F '<fields omitted, see original request>'"
+		curl += "-F '<fields omitted, see original request>' "
 	} else if req.RawRequest.GetBody != nil {
 		// Handle normal body
 		body, err := req.RawRequest.GetBody()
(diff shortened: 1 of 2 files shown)


jeevatkm (commented):
@c-tonneslan, Thanks for the PR. The main branch is not the correct one. Please use v3 for your PR. Many changes are already present in the v3 branch.

jeevatkm (changes requested):

c-tonneslan:
Thanks, retargeted to v3. The fix applies cleanly there since `curl.go` and `curl_test.go` have the same buggy lines on both branches. Tests still pass locally on v3 (`go test ./... -run TestCurl` green).

jeevatkm (approved):
@c-tonneslan Thanks for retargeting to v3 and the fix.

PR #1172, opened 2026-06-11
fix: panics and state corruption in Request.Clone, Client.Close, SSE default logger, and invalid HTTP method handling

g3m0sis:
### 1. `Request.Clone` produces a corrupted cookies slice (`request.go`)

The cookies slice was created with `make([]*http.Cookie, l)` and then appended to, producing a slice of length `2l` whose first half is `nil` entries:

```go
rr.Cookies = make([]*http.Cookie, l)        // length l, all nil
for _, cookie := range r.Cookies {
    rr.Cookies = append(rr.Cookies, cloneCookie(cookie))  // appends after the nils
}
```

Executing the cloned request panics inside `http.Request.AddCookie` on the first `nil` cookie. Fixed by using `make([]*http.Cookie, 0, l)`.

### 2. `Request.Clone` resets the source request instead of the clone (`request.go`)

Three reset statements targeted the receiver `r` instead of the clone `rr`:

```go
r.values = make(map[string]any)   // wipes the parent's values map
r.multipartErrChan = nil
r.ctxCancelFunc = nil
```

This silently wiped the parent request's internal state (e.g., the debug-log entry stored in `values`) and left the clone sharing the parent's multipart error channel and context cancel function.

### 3. Calling `Client.Close` twice panics (`client.go`)

`Close` unconditionally executed `close(c.certWatcherStopChan)`, so a second call panicked with `close of closed channel`. `Close` is now idempotent (guarded by an internal flag under the client lock), and `OnClose` hooks are guaranteed to run exactly once.

`Client.Clone` now also allocates its own `certWatcherStopChan`, so closing both the parent and the clone is safe (previously they shared the channel, and closing both panicked as well).

### 4. `NewSSESource` never initializes its logger (`sse.go`)

The `log` field was left `nil` unless the user explicitly called `SetLogger`. Any code path that logs — overwriting a callback via `OnMessage`/`OnOpen`/`OnError`/`OnRequestFailure`/`AddEventListener`, or a body read failure in `SetBody` — panicked with a nil pointer dereference. The existing tests masked this because the shared test helper always calls `SetLogger`. `NewSSESource` now initializes the default logger, consistent with `createClient`.

### 5. Invalid HTTP method panics instead of returning an error (`middleware.go`)

`MiddlewareRequestCreate` discarded the error returned by `createRawRequest`:

```go
createRawRequest(c, r)  // error ignored
```

The comment assumed all errors were URL-related and already caught in `parseRequestURL`, but `http.NewRequestWithContext` also validates the HTTP method. An invalid method (e.g. `client.R().Execute("BAD METHOD", url)`) left `RawRequest` as `nil` and panicked later in `withTimeout`. The error is now propagated, so the request fails with `net/http: invalid method` as an invalid-request error and triggers the `OnInvalid` hooks.

## Behavior changes (non-breaking)

- `Client.Close()` is now documented and implemented as idempotent; subsequent calls are no-ops and `OnClose` hooks run exactly once. Previously a second call always panicked, so no working code relied on the old behavior.
- `Request.Execute` with an invalid HTTP method now returns an error and invokes `OnInvalid` hooks instead of panicking.

## Tests

- `TestRequestCloneCookiesAndInternalState` — cookies are deep-copied with no `nil` entries; mutating a cloned cookie does not affect the parent; `Clone` no longer resets the parent's `values`/`multipartErrChan`/`ctxCancelFunc`; the clone gets a fresh values map.
- `TestRequestInvalidMethod` — an invalid method returns an `invalid method` error, fires the `OnInvalid` hook, and does not panic.
- `TestClientCloseIdempotent` — double `Close` does not panic and `OnClose` hooks run exactly once.
- `TestClientCloneClose` — closing both the parent client and its clone is safe.
- `TestSSESourceDefaultLogger` — `NewSSESource` has a working default logger; overwriting callbacks and `SetBody` read failures log instead of panicking.

### DIFF
--- a/client.go
+++ b/client.go
@@ -263,6 +263,7 @@ type Client struct {
 	contentDecompresserKeys    []string
 	contentDecompressers       map[string]ContentDecompresser
 	certWatcherStopChan        chan bool
+	isClosed                   bool
 	circuitBreaker             CircuitBreaker
 	hedging                    Hedger
 	rateLimiter                RateLimiter
@@ -2409,11 +2410,22 @@ func (c *Client) Clone(ctx context.Context) *Client {
 
 	// certain values need to be reset
 	cc.lock = &sync.RWMutex{}
+	cc.certWatcherStopChan = make(chan bool)
+	cc.isClosed = false
 	return cc
 }
 
 // Close method executes all registered [CloseHook] callbacks and releases client resources.
+// It is safe to call Close multiple times; subsequent calls are no-op.
 func (c *Client) Close() error {
+	c.lock.Lock()
+	alreadyClosed := c.isClosed
+	c.isClosed = true
+	c.lock.Unlock()
+	if alreadyClosed {
+		return nil
+	}
+
 	// Execute close hooks first
 	c.onCloseHooks()
 
--- a/client_test.go
+++ b/client_test.go
@@ -1719,6 +1719,56 @@ func TestClientCloneClose(t *testing.T) {
 	assertNil(t, cc.Close())
 }
 
+func TestClientCloseConcurrentIdempotent(t *testing.T) {
+	const goroutines = 64
+
+	closeCounter := 0
+	c := dcnl()
+	c.OnClose(func() {
+		// only one goroutine ever reaches the hook, so a plain
+		// increment is data-race free
+		closeCounter++
+	})
+
+	errs := make([]error, goroutines)
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
+	for i := 0; i < goroutines; i++ {
+		go func(idx int) {
+			defer wg.Done()
+			errs[idx] = c.Close()
+		}(i)
+	}
+	wg.Wait()
+
+	// concurrent closes must not panic and every call returns nil
+	for _, err := range errs {
+		assertNil(t, err)
+	}
+	// the OnClose hook runs exactly once
+	assertEqual(t, 1, closeCounter)
+}
+
+func TestClientCloneAfterCloseLifecycle(t *testing.T) {
+	c := dcnl()
+
+	// parent closes first
+	assertNil(t, c.Close())
+	assertEqual(t, true, c.isClosed)
+
+	// a clone taken from a closed parent starts with a fresh lifecycle
+	cc := c.Clone(context.Background())
+	assertEqual(t, false, cc.isClosed)
+
+	// closing the clone is idempotent and must not panic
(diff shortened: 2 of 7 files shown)


jeevatkm (changes requested):
@g3m0sis Thanks for the PR with fixes. I think its good to add couple of more tests inline with changes done.

Can you add following into `client_test.go`?
```go
func TestClientCloseConcurrentIdempotent(t *testing.T) {
	c := dcnl()

	closeCounter := 0
	var closeCounterMu sync.Mutex
	c.OnClose(func() {
		closeCounterMu.Lock()
		closeCounter++
		closeCounterMu.Unlock()
	})

	const closeCallers = 64
	start := make(chan struct{})
	errCh := make(chan error, closeCallers)
	var wg sync.WaitGroup

	for i := 0; i < closeCallers; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			<-start
			errCh <- c.Close()
		}()
	}

	close(start)
	wg.Wait()
	close(errCh)

	for err := range errCh {
		assertNil(t, err)
	}

	closeCounterMu.Lock()
	defer closeCounterMu.Unlock()
	assertEqual(t, 1, closeCounter)
}

func TestClientCloneAfterCloseLifecycle(t *testing.T) {
	parent := dcnl()

	assertNil(t, parent.Close())
	assertTrue(t, parent.isClosed)

	clone := parent.Clone(context.Background())
	assertFalse(t, clone.isClosed)

	cloneCloseCounter := 0
	clone.OnClose(func() {
		cloneCloseCounter++
	})

	assertNil(t, clone.Close())
	assertNil(t, clone.Close())
	assertEqual(t, 1, cloneCloseCounter)

	// parent remains closed and subsequent closes stay no-op
	assertNil(t, parent.Close())
}
```

g3m0sis:
@jeevatkm  I've added the two requested tests in client_test.go:

- TestClientCloseConcurrentIdempotent — 64 goroutines call Close()
  concurrently; asserts no panic, every call returns nil, and the
  OnClose hook runs exactly once.
- TestClientCloneAfterCloseLifecycle — parent closes first, a clone
  taken from the closed parent starts a fresh lifecycle, the clone
  closes idempotently, and the parent stays closed throughout.

Both pass locally under `go test -race`. The CI run is waiting for your
approval to run. Thanks for the review!

jeevatkm (approved):
@g3m0sis Thanks for the PR update.

Issue #1176, opened 2026-06-24
Context cancellation leak


vadimi:
Hello, I've been troubleshooting some memory leaks in one of my services and noticed that resty v3 is leaking/not calling context cancellation methods on the success path. Here is one of the profiles:

<img width="1584" height="291" alt="Image" src="https://github.com/user-attachments/assets/10185a86-0479-43e7-b0cc-10a10f2345ba" />

vadimi:
sent a PR with a potential fix, tested it on my local and I don't see cancellation leaks anymore

vadimi:
@jeevatkm thanks for merging the fix quickly, would you mind pushing a tag?

jeevatkm:
@vadimi I will create a tag as soon as I merge PR is #1178.

jeevatkm:
Fixed by #1177

vadimi:
thanks @jeevatkm, seems like all PRs are resolved/merged. Would you mind pushing a tag please? 🙏

jeevatkm:
@vadimi I'm sorry for the delay. Tonight I will get it done.

jeevatkm:
@vadimi v3.0.0-rc.3 is released

vadimi:
thank you @jeevatkm

PR #1177, opened 2026-06-24
fix context cancellation leak

vadimi:
On the success path the request-level timeout context's cancel func is never called, so it leaks

related to https://github.com/go-resty/resty/issues/1176

### DIFF
--- a/client.go
+++ b/client.go
@@ -2490,10 +2490,21 @@ func (c *Client) execute(req *Request) (*Response, error) {
 		req.multipartCancelFunc()
 	}
 
+	// Take ownership of the per-attempt timeout cancel func set by
+	// withTimeout. It must fire once the response body is fully consumed,
+	// so it is attached to the body's Close below. On a transport error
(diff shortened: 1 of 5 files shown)


jeevatkm (changes requested):
@vadimi Thanks for the PR.

Can you please add test cases for patch coverage? See codecov report.

vadimi:
@jeevatkm of course, done

jeevatkm:
Thanks @vadimi, It seems CI run failed in `go fmt` validation. Can you please check?

jeevatkm:
Thanks @vadimi, there is a tiny difference in patch coverage. It seems this use case flow is not covered.
```go
// client.go 

// No response body was available to attach the cancel to (e.g. resp is
+ 	// nil); release the timeout context now so it does not leak.
+ 	if cancel != nil {
+ 		cancel()
+ 	}
```
I will merge this PR. I will take care later on if you can feel free.

PR #1178, opened 2026-06-25
fix(retry): protect parseRetryAfterHeader from integer and time duration overflow

HarshalPatel1972:
### Description
This PR addresses an un-guarded integer and time duration overflow vulnerability within the automatic retry header parsing engine (`parseRetryAfterHeader`). By introducing safe multiplication boundary thresholds, we ensure the client cleanly caps astronomical server-provided delays instead of corrupting the backoff context scheduler or causing a runtime panic.

### Technical Context
When parsing an HTTP `Retry-After` header value, the client evaluates either delta-seconds or a specific HTTP-Date string. 
* In the seconds track, multiplying an arbitrary 64-bit integer directly by `time.Second` ($1,000,000,000$ nanoseconds) easily overflows the internal maximum limit of a Go `time.Duration` structure ($2^{63}-1$). This calculation error wraps the resulting delay into a negative duration, causing sub-system scheduling panics.
* In the date fallback track, subtracting the current epoch time from an extreme future date parameter creates a differential calculation that can similarly exceed duration storage limits.

### Resolution Strategy
* **Multiplication Ceiling Fencing:** Injected an inline arithmetic boundary check using `math.MaxInt64 / time.Second` inside `retry.go` to intercept excessive delay parameters out of the gate.
* **Duration Range Normalization:** Forced the computed date subtraction result to validate within expected positive limits, defaulting seamlessly to maximum bounds if a range wrap or negative anomaly is encountered.

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -222,6 +222,9 @@ func parseRetryAfterHeader(v string) (time.Duration, bool) {
 		if delay < 0 { // a negative delay doesn't make sense
 			return 0, false
 		}
+		if delay > int64(math.MaxInt64/time.Second) {
+			return time.Duration(math.MaxInt64), true
+		}
 		return time.Second * time.Duration(delay), true
 	}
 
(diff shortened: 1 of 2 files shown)


jeevatkm (changes requested):
@HarshalPatel1972 Thanks for the PR. 

Can you add test cases for patch coverage? See codecov report.

jeevatkm:
@HarshalPatel1972 ping

HarshalPatel1972:
> @HarshalPatel1972 ping

@jeevatkm Good catch! Just pushed an update to fix the coverage.

I added a couple of test cases in retry_test.go to specifically hit those overflow paths. One passes an intentionally massive integer to hit the seconds ceiling, and the other uses a date way out in the year 2293 to trigger the duration wrap-around.

Tests are passing locally, so codecov should be happy now. Let me know if everything looks good on your end!

jeevatkm:
@HarshalPatel1972 Thanks for updating tests. It seems the overflow path is not covered. Can you please check codecov?

jeevatkm (approved):
@HarshalPatel1972 I have added a test case and merging it.

PR #1179, opened 2026-06-28
add unit test to cover cancellation on nil resp body

vadimi:
adding missing unit test - follow up to https://github.com/go-resty/resty/pull/1177#issuecomment-4821683041


jeevatkm (approved):
Thanks @vadimi for the follow-up to take care tests.

PR #1180, opened 2026-07-03
improve string request body parsing memory allocs

vadimi:
`bodyBuf.Write([]byte(body))` allocates memory for `[]byte`. There is `WriteString` method on `bytes.Buffer` that accepts a string and does not allocate.


jeevatkm (approved):
Thanks @vadimi for the PR.

PR #1185, opened 2026-07-17
docs(sse): replace dead sse.dev example host with a live public SSE endpoint

donbagger:
## Summary

The SSE godoc examples in `sse.go`, plus one URL fixture in `sse_test.go`, point to `https://sse.dev/test`. That domain is gone:

```
$ host sse.dev
Host sse.dev not found: 3(NXDOMAIN)
```

Anyone copy-pasting the examples gets a DNS error instead of a stream.

This PR swaps in a live public SSE endpoint that needs no signup or API key (DexPaprika WETH price stream):

```
https://streaming.dexpaprika.com/sse/prices?method=token_price&chain=ethereum&address=0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
```

It emits named `token_price` events when the tracked price updates, which for WETH works out to one every several seconds. Because the events are named, I also switched the two full examples from `OnMessage` to `AddEventListener("token_price", ...)`. `OnMessage` only fires for unnamed (`message`) events, so against a named-event stream the old example shape compiles but prints nothing. I ran both variants against this branch to confirm: the listener receives events, `OnMessage` receives zero.

The `sse_test.go` change is only the URL string in `TestSSESourceCoverage`. That test never dials (Get() fails on the "at least one OnMessage/AddEventListener" validation first), so this just removes the last reference to the dead host.

## Testing

- `diff -u <(echo -n) <(go fmt $(go list ./...))` is clean
- `go test ./... -race -count=1 -covermode=atomic -coverpkg=./... -shuffle=on` passes, coverage unchanged at 99.9%
- Ran the updated example end to end against the live endpoint through this branch's SSE implementation and confirmed event delivery

For transparency: I do DevRel for DexPaprika, which is how I knew about this endpoint. It's free and keyless, and it's meant to stay up long-term as a public example host. If you'd rather point the examples at a different public SSE host, happy to update.

### DIFF
--- a/sse.go
+++ b/sse.go
@@ -97,8 +97,8 @@ type (
 // NewSSESource creates a new [SSESource] with default SSE settings.
 //
 //	sse := NewSSESource().
-//		SetURL("https://sse.dev/test").
-//		OnMessage(
+//		SetURL("https://streaming.dexpaprika.com/sse/prices?method=token_price&chain=ethereum&address=0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2").
+//		AddEventListener("token_price",
 //			func(e any) {
 //				event := e.(*resty.SSE)
 //				fmt.Println(event)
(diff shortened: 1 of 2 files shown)


jeevatkm:
@donbagger Thanks for reporting the see.dev domain issue. I'm sorry, I'm unable to accept this PR due to business-related references. I will figure out an appropriate resolution.

donbagger:
Understood, and thanks for saying so directly rather than leaving it open.

I have opened #1205 with no business reference in it. It only removes the dead host and puts `https://example.com/events` in its place, which is what the rest of the package already uses for illustrative URLs. Four lines, `OnMessage` untouched.

Closing this one. `sse.dev` is still NXDOMAIN as of today, so the docs issue is real regardless of which replacement you prefer, and if a placeholder is not what you want I offered an `httptest`-based example on the new PR instead.

PR #1187, opened 2026-07-21
feat: add support for the HTTP QUERY method (RFC 10008)

thatwasyahya:
## What

[RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html) (Proposed Standard, June 2026) defines **QUERY**, a **safe** and **idempotent** HTTP method that carries the query as request content — combining GET's safety, cacheability and retryability with POST's arbitrary request body. This PR adds first-class support for it to Resty.

## Changes

- Add the `MethodQuery` constant (`client.go`).
- Add the `Request.Query(url)` helper, matching the existing verb helpers such as `Get`/`Post`/`Trace` (`request.go`).
- Send the request body for QUERY in `isPayloadSupported()` — QUERY always carries content, so this is required for the method to be usable (`request.go`).
- Classify QUERY as **idempotent** (so automatic retries apply) and **read-only** (so hedging applies), consistent with its RFC 10008 safe + idempotent semantics (`request.go`, `hedging.go`).

## Ecosystem context

QUERY is registered in the IANA HTTP Method Registry as safe + idempotent (RFC 10008 Section 5.1), and the Go router ecosystem is already adopting it — **chi** (`Router.Query`) and **echo** (`Echo.QUERY`) both ship it. This brings the same first-class support to Resty's client.

## Tests

- `TestQuery` — end-to-end `Query()` call with a body against a test server.
- `TestQueryMethodClassification` — asserts QUERY is treated as read-only and idempotent.

`go test ./...` passes.

## Notes

Happy to split this up or adjust naming if you'd prefer.

### DIFF
--- a/request.go
+++ b/request.go
@@ -1438,6 +1438,14 @@ func (r *Request) Trace(url string) (*Response, error) {
 	return r.Execute(MethodTrace, url)
 }
 
+// Query method does QUERY HTTP request. QUERY is a safe, idempotent method that
+// carries the query as request content; it's defined in section 2 of [RFC 10008].
+//
+// [RFC 10008]: https://datatracker.ietf.org/doc/html/rfc10008.html#section-2
+func (r *Request) Query(url string) (*Response, error) {
+	return r.Execute(MethodQuery, url)
+}
+
 // Send method performs the HTTP request using the method and URL already defined
 // for current [Request].
 //
@@ -1800,7 +1808,8 @@ func (r *Request) isPayloadSupported() bool {
 		return true
 	}
 
-	if r.Method == MethodPost || r.Method == MethodPut || r.Method == MethodPatch {
+	// QUERY (RFC 10008) always carries the query as request content.
+	if r.Method == MethodPost || r.Method == MethodPut || r.Method == MethodPatch || r.Method == MethodQuery {
 		return true
 	}
 
@@ -1853,6 +1862,7 @@ var idempotentMethods = map[string]struct{}{
 	MethodHead:    {},
 	MethodOptions: {},
 	MethodPut:     {},
+	MethodQuery:   {},
 	MethodTrace:   {},
 }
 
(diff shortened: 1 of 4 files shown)


request.go
```diff
@@ -1438,6 +1438,14 @@ func (r *Request) Trace(url string) (*Response, error) {
 	return r.Execute(MethodTrace, url)
 }
 
+// Query method does QUERY HTTP request. QUERY is a safe, idempotent method that
+// carries the query as request content; it's defined in section 2 of [RFC 10008].
+//
+// [RFC 10008]: https://datatracker.ietf.org/doc/html/rfc10008.html#section-2
+func (r *Request) Query(url string) (*Response, error) {
```
jeevatkm, line 1449: @thatwasyahya Thanks for the PR. I appreciate your interest in the adoption of HTTP `Query` method. It is just a month old and in the very early adopter phase in the industry.

Could you please add the following `NOTE` in the godoc?

```
// NOTE:
//   - RFC is in the early stage of adoption in the industry; server side and browsers 
//     are yet to add support. You may run into unexpected failures due to infrastructure/environment.
```

Also, please run the following command and update the PR.
```
$ gofmt -w .
```

jeevatkm (changes requested):
@thatwasyahya Thanks for the PR.

jeevatkm:
@thatwasyahya The codecov coverage report didn't show up on the PR. I will troubleshoot over the weekend. For now, could you please run the following commands to ensure code coverage?

```
# Test and generate report

# to view coverage report in HTML
```

thatwasyahya:
Thanks for the review, @jeevatkm! Addressed:

- Added the early-adoption **NOTE** to the `Query` godoc, as suggested.
- Ran `gofmt -w .` — the tree is clean.
- Ran the coverage command you provided:

  ```sh
  go test ./... -race -count=1 -coverprofile=coverage.txt -covermode=atomic -shuffle=on -coverpkg=./...
  ```

  Result: **99.9%** overall, and the new code is fully covered — `Query`, `isPayloadSupported`, `isIdempotent`, and `isReadOnlyMethod` are all at **100.0%** (`go tool cover -func`).

Pushed as a follow-up commit. Happy to squash or adjust anything else.

jeevatkm (approved):
@thatwasyahya Thanks for updating the PR.

PR #1188, opened 2026-07-21
fix: stop multipart producer after early response without race/hang

### LINKED ISSUE #1186: v3: multipart upload has a data race and can hang after an early server response
## Summary

Resty v3 multipart streaming has two related concurrency problems when the HTTP server or `RoundTripper` returns a response before the multipart producer has finished writing the request body:

1. `Request.multipartCancelFunc` is written by the multipart writer goroutine and read/called by the request goroutine without synchronization.
2. The request can hang indefinitely while waiting for `multipartErrChan` if the source reader is already blocked inside `Read`.

This appears to be related to the multipart lifecycle introduced in #1101 while fixing #1030.

## Versions tested

- `resty.dev/v3 v3.0.0-rc.3`
- current `v3` branch, commit `ea5b6cf677568d2f22b395a9dca8b95dec43f939`
- Go 1.25.4
- darwin/arm64

Both problems reproduce on `v3.0.0-rc.3` and the current `v3` branch.

## Reproducer

`go.mod`:

```go
module resty-multipart-repro

go 1.23

require resty.dev/v3 v3.0.0-rc.3
```

`multipart_test.go`:

```go
package restyrepro

import (
	"bytes"
	"io"
	"net/http"
	"net/http/httptest"
	"strings"
	"sync"
	"testing"
	"time"

	"resty.dev/v3"
)

func TestMultipartEarlyResponseRace(t *testing.T) {
	client := resty.NewWithClient(&http.Client{
		Transport: earlyResponseTransport{},
	})

	for range 1000 {
		_, _ = client.R().
			SetMultipartFields(&resty.MultipartField{
				Name:        "audio",
				FileName:    "audio.wav",
				ContentType: "audio/wav",
				Reader:      bytes.NewReader(make([]byte, 1<<20)),
			}).
			Post("http://resty.test/upload")
	}
}

type earlyResponseTransport struct{}

func (earlyResponseTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	// net/http allows a RoundTripper to consume and close the request body
	// asynchronously after RoundTrip returns.
	go func() {
		_, _ = io.Copy(io.Discard, req.Body)
		_ = req.Body.Close()
	}()

	return &http.Response{
		StatusCode: http.StatusUnprocessableEntity,
		Status:     "422 Unprocessable Entity",
		Header:     make(http.Header),
		Body:       io.NopCloser(strings.NewReader("rejected")),
		Request:    req,
	}, nil
}

type blockingReadCloser struct {
	started chan struct{}
	release chan struct{}
	once    sync.Once
}

func newBlockingReadCloser() *blockingReadCloser {
	return &blockingReadCloser{
		started: make(chan struct{}),
		release: make(chan struct{}),
	}
}

func (r *blockingReadCloser) Read([]byte) (int, error) {
	select {
	case <-r.started:
	default:
		close(r.started)
	}

	<-r.release

	return 0, io.EOF
}

func (r *blockingReadCloser) Close() error {
	r.once.Do(func() {
		close(r.release)
	})

	return nil
}

func TestMultipartReturnsAfterEarlyResponse(t *testing.T) {
	reader := newBlockingReadCloser()

	server := httptest.NewServer(http.HandlerFunc(
		func(w http.ResponseWriter, _ *http.Request) {
			// Ensure the multipart writer is blocked inside the source
			// Reader.Read before returning an early response.
			<-reader.started
			http.Error(w, "rejected", http.StatusUnprocessableEntity)
		},
	))
	defer server.Close()

	done := make(chan error, 1)

	go func() {
		_, err := resty.New().R().
			SetMultipartFields(&resty.MultipartField{
				Name:        "audio",
				FileName:    "audio.wav",
				ContentType: "audio/wav",
				Reader:      reader,
			}).
			Post(server.URL)

		done <- err
	}()

	select {
	case <-done:
		// The HTTP response has already arrived, so the request should
		// return without requiring the caller to unblock the reader.
	case <-time.After(time.Second):
		// Cleanup the blocked Resty multipart writer.
		_ = reader.Close()
		<-done

		t.Fatal("multipart request did not return after the server responded")
	}
}
```

## Data race

Run:

```shell
go test -race -run TestMultipartEarlyResponseRace -count=1
```

The race detector reports:

```text
WARNING: DATA RACE

Read at ... by request goroutine:
  resty.dev/v3.(*Client).execute()
      client.go:2489

Previous write at ... by multipart writer goroutine:
  resty.dev/v3.handleMultipart.func1()
      middleware.go:403
```

The relevant accesses are:

- assignment from the multipart writer goroutine:

  https://github.com/go-resty/resty/blob/v3/middleware.go#L402-L403

- read and invocation from `Client.execute`:

  https://github.com/go-resty/resty/blob/v3/client.go#L2488-L2491

The custom `RoundTripper` follows the `net/http.RoundTripper` contract. The standard library documentation explicitly permits closing the request body in a separate goroutine after `RoundTrip` returns:

https://pkg.go.dev/net/http#RoundTripper

## Hang

Run:

```shell
go test -run TestMultipartReturnsAfterEarlyResponse -count=3
```

Actual result:

```text
multipart request did not return after the server responded
```

The server has already returned an HTTP 422 response, but `Client.execute` waits for `multipartErrChan` to be closed.

Calling `multipartCancelFunc` does not unblock the request because `gracefulStopReader` checks the context only before calling the underlying reader:

```go
func (gsr *gracefulStopReader) Read(p []byte) (int, error) {
	if err := gsr.ctx.Err(); err != nil {
		return 0, io.EOF
	}
	return gsr.r.Read(p)
}
```

If the source is already blocked inside `gsr.r.Read(p)`, cancelling the context has no effect. The multipart goroutine cannot finish, `multipartErrChan` is not closed, and `Client.execute` waits indefinitely.

Although the source implements `io.Closer`, Resty only closes multipart fields after `Client.execute` returns, which is too late for this situation.

## Expected behavior

- No unsynchronized access to `multipartCancelFunc`.
- A completed HTTP response should not leave the Resty request blocked waiting for the multipart producer.
- A closable source reader should be closed when multipart production needs to be interrupted.
- Context cancellation or request timeout should not leave `Client.execute` waiting indefinitely for `multipartErrChan`.

## Possible direction

Creating the multipart context and cancel function synchronously before starting the writer goroutine would remove the publication race.

The multipart shutdown flow may also need to:

1. cancel the writer context;
2. close the pipe;
3. close source readers that implement `io.Closer`;
4. avoid waiting indefinitely for a producer blocked in an arbitrary `io.Reader.Read`.

A mutex around `multipartCancelFunc` alone would address the race but would not fix the hang.

Solaris-star:
## Summary
Fixes #1186.

Resty v3 multipart streaming had two related problems when a `RoundTripper` (or server) returned a response before the multipart producer finished writing the request body:

1. **Data race** on `Request.multipartCancelFunc` — written in the multipart writer goroutine and read/called from `Client.execute` without synchronization.
2. **Hang** waiting on `multipartErrChan` when the source `io.Reader` was already blocked inside `Read`, because context cancel alone could not unblock it.

## Fix
- Create the multipart cancel context **before** starting the writer goroutine (publish-before-start).
- After `http.Client.Do` returns, call `stopMultipart()` which:
  - cancels the multipart context
  - closes the pipe writer
  - closes closable multipart field readers
- Treat intentional shutdown signals (`io.ErrClosedPipe` / context cancel) as non-failures when an HTTP response was already obtained, so successful early responses are not turned into `CascadeError`.

## Test plan
- [x] `go test -race -run 'Multipart|multipart' -count=1`
- [x] `TestMultipartEarlyResponseRace`
- [x] `TestMultipartReturnsAfterEarlyResponse`

### DIFF
--- a/client.go
+++ b/client.go
@@ -2485,9 +2485,12 @@ func (c *Client) execute(req *Request) (*Response, error) {
 
 	req.StartTime = time.Now()
 	resp, err := c.Client().Do(req.withTimeout())
-	// Cancel multipart context for io.Copy to stop reading/writing further
-	if req.isMultiPart && req.multipartCancelFunc != nil {
-		req.multipartCancelFunc()
+	// Stop multipart production once the transport has returned. Cancel alone
+	// is not enough when a source Reader is already blocked inside Read; also
+	// close closable field readers so the writer goroutine can finish and
+	// multipartErrChan can close (see #1186).
+	if req.isMultiPart {
+		req.stopMultipart()
 	}
 
 	// Take ownership of the per-attempt timeout cancel func set by
@@ -2509,8 +2512,14 @@ func (c *Client) execute(req *Request) (*Response, error) {
 	}
 	if req.isMultiPart && req.multipartErrChan != nil {
 		// read all multipart errors from channel
-		for err = range req.multipartErrChan {
-			response.CascadeError = wrapErrors(err, response.CascadeError)
+		for merr := range req.multipartErrChan {
+			// stopMultipart intentionally closes the pipe/readers after the
+			// transport returns. Those shutdown signals are not request failures
+			// when an HTTP response was already obtained.
+			if resp != nil && isMultipartStopError(merr) {
+				continue
+			}
+			response.CascadeError = wrapErrors(merr, response.CascadeError)
 		}
 	}
 
--- a/middleware.go
+++ b/middleware.go
@@ -373,6 +373,7 @@ func handleMultipart(c *Client, r *Request) error {
 	br, bw := io.Pipe()
 	mw := multipart.NewWriter(bw)
 	r.Body = br
+	r.multipartPipeWriter = bw
 
 	// set custom multipart boundary if exists
 	if err := multipartSetBoundary(mw, r); err != nil {
@@ -382,6 +383,10 @@ func handleMultipart(c *Client, r *Request) error {
 
 	r.Header.Set(hdrContentTypeKey, mw.FormDataContentType())
 
+	// Create cancel before the writer goroutine so Client.execute can stop
+	// production without racing on multipartCancelFunc publication.
+	ctx, cancel := context.WithCancel(r.Context())
+	r.multipartCancelFunc = cancel
 	r.multipartErrChan = make(chan error, 1)
 	go func() {
 		defer close(r.multipartErrChan)
@@ -399,8 +404,6 @@ func handleMultipart(c *Client, r *Request) error {
 			return
 		}
 
-		ctx, cancel := context.WithCancel(r.Context())
-		r.multipartCancelFunc = cancel
 		for _, mf := range r.multipartFields {
 			if mf.isValues() {
 				for _, v := range mf.Values {
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 )
 
 func Test_parseRequestURL(t *testing.T) {
@@ -1096,6 +1097,7 @@ func TestMiddleware_multipartWriteFormData(t *testing.T) {
 	})
 
 	req := &Request{
+		mu:          new(sync.Mutex),
 		Header:      http.Header{},
 		isMultiPart: true,
 		multipartFields: []*MultipartField{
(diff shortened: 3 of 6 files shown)


jeevatkm:
@Solaris-star  Thanks for the PR. 

The codecov coverage report didn't show up on the PR. I will troubleshoot over the weekend. For now, could you please run the following commands to ensure code coverage?

```
# Test and generate report

# To view coverage report in HTML
```

Solaris-star:
Thanks @jeevatkm — ran the coverage commands you asked for on this branch:

```bash
go test ./... -race -count=1 -coverprofile=coverage.txt -covermode=atomic -shuffle=on -coverpkg=./...
```

Result:

- **statement coverage: 99.7%** of `./...`
- Codecov on the PR is also reporting project coverage ~99.68% (patch ~86%)

Two existing-looking failures showed up in this environment (`TestTraceInfoOnTimeout` / `TestTraceInfoOnTimeoutWithSetTimeout`) that appear unrelated to the multipart cancel path; happy to dig further if you want those green in CI as a hard gate too.

jeevatkm:
@Solaris-star Thanks for getting back. Could you please make patch coverage as close to 100% as possible?

Solaris-star:
Thanks @jeevatkm — pushed a follow-up that brings the new helpers to 100% patch coverage:

- `stopMultipart`: added a nil-receiver guard test
- `isMultipartStopError`: added a table test covering `nil`, the sentinel errors (`io.ErrClosedPipe`, `context.Canceled`, `context.DeadlineExceeded`), a wrapped error, and the string-fallback matches

Verified locally:

```
resty.dev/v3/request.go:1604: stopMultipart          100.0%
resty.dev/v3/util.go:550:     isMultipartStopError   100.0%
```

(The two `TestTraceInfoOnTimeout*` failures you may see in CI reproduce on a clean checkout without my change — they're the environment-dependent timing flakes already noted in #1184, unrelated to this PR.)

jeevatkm:
@Solaris-star Thanks for the update. Currently, it’s failing due to a Go format file error. Please see the CI build log.

Solaris-star:
Thanks — I ran `gofmt` on `util_test.go` and pushed the formatting-only follow-up in `b2dfe5b`. The focused multipart tests pass locally, and `gofmt -l util_test.go` is now empty.

PR #1192, opened 2026-08-02
fix: handle Windows connection refused load balancer feedback

deepakganesh78:
## Summary

Treat Windows `WSAECONNREFUSED` (`syscall.Errno(10061)`) the same as `syscall.ECONNREFUSED` when Resty reports request feedback to a load balancer.

## Reproduction

On Windows, a request routed by `WeightedRoundRobin` to a closed backend returns an error chain containing `*os.SyscallError` with `syscall.Errno(10061)`. `errors.Is(err, syscall.ECONNREFUSED)` does not match that errno, so `sendLoadBalancerFeedback` reports the attempt as successful and the backend never reaches `MaxFailures`.

## Root cause

The load-balancer feedback path only recognized `syscall.ECONNREFUSED`. Windows uses Winsock error 10061 (`WSAECONNREFUSED`) for the same condition, so connection-refused failures were ignored on Windows.

## Fix

Add a small `isConnectionRefused` helper that recognizes both the portable `syscall.ECONNREFUSED` value and Windows `WSAECONNREFUSED`, and cover the behavior with `TestLoadBalancerConnectionRefusedMarksHostInactive`.

## Compatibility

No API changes and no new dependencies. Non-Windows behavior remains unchanged.

## Validation

- `go test -run "TestLoadBalancerConnectionRefusedMarksHostInactive|TestLoadBalancerRequestFailures" -count=1` — PASS
- Regression check: `TestLoadBalancerConnectionRefusedMarksHostInactive` fails on the previous implementation on Windows (`Expected [0], got [1]`).
- `go build ./...` — PASS
- `go vet ./...` — PASS
- `gofmt -l .` — empty
- `go test ./...` — attempted 3 times on Windows; this fix removes the prior `TestLoadBalancerRequestFailures` failure, but the full suite still has unrelated/pre-existing Windows issues: `TestTraceInfo/enable_trace_on_invalid_request,_issue_#1016` observes `TotalTime == 0` for an immediate invalid scheme, `TestOutputPathDirNotExists` intermittently fails, and `TestClient_SetRootCertificateWatcher` once hit a Windows file-lock error deleting `root-ca.crt`.

### DIFF
--- a/request.go
+++ b/request.go
@@ -19,6 +19,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 	"sync"
 	"syscall"
@@ -1857,7 +1858,7 @@ func (r *Request) sendLoadBalancerFeedback(res *Response, err error) {
 	if err != nil {
 		var noe *net.OpError
 		if errors.As(err, &noe) {
-			success = !errors.Is(noe.Err, syscall.ECONNREFUSED) || noe.Timeout()
+			success = !isConnectionRefused(noe.Err) || noe.Timeout()
 		}
 	}
 	if success && res != nil &&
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -1872,6 +1873,18 @@ func (r *Request) sendLoadBalancerFeedback(res *Response, err error) {
 	})
 }
 
+const windowsWSAECONNREFUSED syscall.Errno = 10061
+
+func isConnectionRefused(err error) bool {
+	if errors.Is(err, syscall.ECONNREFUSED) {
+		return true
+	}
+	if runtime.GOOS == "windows" && errors.Is(err, windowsWSAECONNREFUSED) {
```
jeevatkm, line 1882: @deepakganesh78 I think OS check is not required. Please let me know if it’s strongly required.

jeevatkm (changes requested):
@deepakganesh78 Thanks for the PR. Can you please check the codecov report and make patch coverage 100%?

jeevatkm (approved):
@deepakganesh78 Thanks for the PR.

PR #1197, opened 2026-08-10
fix(enhancement): hedging no longer cancels the winning response body

mrueg:
`Hedging.RoundTrip` derived every hedged attempt from one shared context and cancelled it as soon as a winner was known — eagerly inside `once.Do`, and again via `defer` on return. That context belongs to the attempt whose `*http.Response` is handed back, so the transport tore the connection down before the caller could read the body.

Anything not already sitting in the transport's read buffer was lost. Measured against a 512 KB response:

| | bytes read | error |
|---|---|---|
| before | 3,965 | `context canceled` |
| after | 524,288 | none |

## Fix

Each attempt gets its own context derived from the request context. The first to finish wins and cancels the others immediately — same behaviour as before, so losing attempts are still abandoned right away — but the winner's cancel func is attached to its response body, so the context lives exactly as long as the caller needs it.

Also in the same function:

- `maxRequest`, `delay`, `rateDelay`, `isNonReadOnlyAllowed` and `underlying` are now read under the lock. `RoundTrip` read three of them unsynchronised while the corresponding setters write under it.
- `break` inside a `select` breaks the select, not the loop, so a cancelled context did not stop further attempts from being spawned. Now a labelled break.
- Dropped the deadline branch: `WithCancel(ctx)` already inherits `ctx`'s deadline, so `WithTimeout(ctx, time.Until(deadline))` was equivalent.

## Why this wasn't caught

The existing hedging tests assert on an 11-byte body, which fits in the transport's read buffer and so survives the early cancel. The two added tests use a 512 KB body — one via `Response.Bytes`, one via `SetResponseDoNotParse` with a delayed read. Both fail on `v3` and pass here.

## Verification

This is one of four PRs from an audit of the v3 branch; the others are independent and touch different files.

### DIFF
--- a/hedging.go
+++ b/hedging.go
@@ -19,7 +19,7 @@ import (
 // Implementations must also implement [http.RoundTripper] so they can be installed
 // as the HTTP transport on a [Client].
 //
-// The [SetTransport] and [Transport] methods allow [Client.SetHedging] to wrap
+// The [Hedger.SetTransport] and [Hedger.Transport] methods allow [Client.SetHedging] to wrap
 // and unwrap the underlying transport when hedging is enabled or disabled.
 //
 // Use [NewHedging] to create the default implementation ([Hedging]).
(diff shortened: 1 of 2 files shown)


jeevatkm (changes requested):
@mrueg Thanks for the PR. Can you please cover patch coverage? see codecov report.

jeevatkm (approved):

PR #1199, opened 2026-08-10
fix: SSE connect deadlock from a recursive read lock, plus a leaked response body

mrueg:
## Deadlock

`SSESource.connect` held `sse.lock.RLock()` for its entire body — across `httpClient.Do`, the retry wait, and the nested `RLock` acquisitions in `createRequest` → `Context()` and in `triggerOnRequestFailure`.

`sync.RWMutex` read locks are **not reentrant**. A concurrent `Close` or setter taking the write lock between the outer and the inner `RLock` blocks both permanently: the writer waits on `connect`'s read lock, and `connect`'s nested read lock queues behind the waiting writer. Any application calling `Close`, `SetHeader` or `SetContext` while a connection is being established could hang forever.

`connect` now snapshots the settings it needs under one short read lock and does its I/O without holding it. `createRequest` resolves the context before taking the lock, for the same reason.

The added test starts `Get()` against a server that returns 429 (so `connect` retries), hammers `SetHeader` from another goroutine, and fails with a message rather than hanging. On `v3` it deadlocks:

```
--- FAIL: TestSSESourceConcurrentCloseDoesNotDeadlock (10.01s)
    SSESource.Get did not return; connect deadlocked against a concurrent setter
```

## Leaked body, and a lost status code

Two further problems in the same function:

- A response not returned to the caller was never drained or closed. The non-retryable path released it only if an `OnRequestFailure` callback was registered; the retry-exhausted path never did.
- When the final attempt yielded a non-2xx response with no transport error, `err = doErr` left `err` nil, so `connect` fell through to the generic `resty:sse: unable to connect stream` and the status code was lost:

```
before: resty:sse: unable to connect stream
after:  resty:sse: 429 Too Many Requests
```

The failure callback still receives the body first, as documented; the body is released only after it returns.

## Also

- `SetURL` and `SetMethod` wrote `sse.url` / `sse.method` without the lock, unlike every other setter, and `Get` read and wrote them unlocked. `Get` now validates and resets in one critical section.
- `parseEventFunc` appended the `\n` separator onto a slice aliasing `bufio.Scanner`'s buffer, which can write over bytes the scanner has not yet handed out. It appends into the event buffer instead. Benign today only because the byte it overwrites happens to be the delimiter.
- Sentinel `err == io.EOF` replaced with `errors.Is`.

## Verification

One of four independent PRs from an audit of the v3 branch.

### DIFF
--- a/sse.go
+++ b/sse.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -138,6 +139,8 @@ func NewSSESource() *SSESource {
 //
 //	sse.SetURL("https://example.com/events")
 func (sse *SSESource) SetURL(url string) *SSESource {
+	sse.lock.Lock()
+	defer sse.lock.Unlock()
 	sse.url = url
 	return sse
 }
@@ -146,6 +149,8 @@ func (sse *SSESource) SetURL(url string) *SSESource {
 //
 //	sse.SetMethod("POST"), or sse.SetMethod(resty.MethodPost)
 func (sse *SSESource) SetMethod(method string) *SSESource {
+	sse.lock.Lock()
+	defer sse.lock.Unlock()
 	sse.method = method
 	return sse
 }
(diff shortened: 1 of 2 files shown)


jeevatkm (changes requested):
@mrueg Thanks for the PR. Can you please cover patch coverage? See codecov report.

jeevatkm (approved):

PR #1200, opened 2026-08-10
fix: three panics, a data race, and a broken multipart retry path

mrueg:
> **Rebased onto #1212** (2026-09-07) so it no longer conflicts with the naming changes there. This branch now sits on top of #1208 → #1209 → #1210 → #1211 → #1212 and shows their commits until they merge; its own delta is unchanged at 10 files, +293/−22. It was also 5 commits behind `v3`, so this brings it current.
>
> Two things merged rather than picked during the rebase:
> - `AddRequestMiddleware` is renamed to `AddRequestMiddlewares` (variadic) in #1212. The `max()` guard against the negative insert index from this PR is preserved in the renamed function, so the panic fix survives either merge order.
> - `Client.execute` now uses both fixes at once: the `c.CircuitBreaker()` accessor added here, and the unexported hook-dispatch interface from #1212.
>
> The commit body below still refers to `Client.AddRequestMiddleware` by its pre-rename name; that describes the bug as it was, and I left the original wording alone.

Five small independent defects, each with a regression test in the matching test file. Grouped because each is a one- or two-line change; happy to split if you'd rather review them separately.

## Panics

**digest — nil map entry.** `createCredentials` calls `newHashFunc(cha.algorithm)` to hash the body for `qop=auth-int`, but the algorithm is only validated later, in `digestCredentials.digest`. `digestHashFuncs[unknown]` is a nil func value, so a server offering an unsupported algorithm together with `auth-int` crashes the process from inside `RoundTrip`. Validated up front; `newHashFunc` no longer calls a nil entry.

**circuit breaker — integer divide by zero.** `slidingWindow.AddAndGet` computes `interval / len(values)`, which is 0 when `resetTimeout` is 0, then divides `elapsed` by it. `NewCircuitBreakerCount(5, 1, 0)` panics on the first request. A non-positive `resetTimeout` now defaults to 30s, `AddAndGet` guards the zero case, and the constructors clamp their other non-positive arguments to the values they already behaved as.

**`Client.AddRequestMiddleware` — negative slice index.** `idx := len(c.beforeRequest) - 1` is `-1` after `SetRequestMiddlewares()` with no arguments, and `slices.Insert` panics on a negative index. Clamped. Its godoc also said "appends", while it inserts *before* the final middleware so user middleware runs ahead of `MiddlewareRequestCreate` — documented as it behaves.

## Data races

**`Client.circuitBreaker`.** `Client.execute` and `cbRequestError` read it directly while `SetCircuitBreaker` writes it under the write lock. Added the accessor the equivalent fields already have (`RateLimiter`, `Hedging`, `LoadBalancer`). The added test reports **64 races** under `-race` without this change.

**`clientTrace.endTime`.** `Response.setReceivedAt` wrote it with no lock, while `Request.TraceInfo` reads it under `ct.lock.RLock` and every other `clientTrace` field is written under that lock. No regression test for this one: reaching it means touching a `Request` while it is in flight, which also races on `r.trace` and `r.Attempt` because a `Request` isn't safe for concurrent use. `TraceInfo`'s godoc now says so.

## Multipart retry

`Request.resetFileReaders` called `resetReader` on every multipart field, including those carrying only `Values`, whose `Reader` is nil. The nil reader fails the `io.ReadSeeker` assertion and returns `ErrReaderNotSeekable`, so **any retry of a request built with `SetMultipartOrderedFormData` was abandoned instead of retried**. Value-only fields are now skipped; the producer rewrites them from `Values` on every attempt.

## Test flake

`TestCircuitBreakerOpenCancelsPreviousResetTimer` sampled `getState()` at a fixed wall-clock instant 30 ms before the half-open transition was due (60 ms reset timeout; 40 ms + 30 ms of sleeps against a 100 ms deadline). A sleep overrun greater than 30 ms reports `HalfOpen` where `Open` was expected — which is the failure I hit once in a full `-race` run.

It now records *when* the transition happens and asserts only that it did not happen sooner than `resetTimeout` after the second `open()`. Timers and sleeps can fire late but never early, so that assertion cannot fail from scheduling delay. Note: I could not reproduce the original flake on demand (105+ runs across CPU saturation and `GOMAXPROCS=1`), so this removes the failure mode by construction rather than on the strength of a before/after reproduction.

## Verification

One of four independent PRs from an audit of the v3 branch.

### DIFF
--- a/cert_watcher_test.go
+++ b/cert_watcher_test.go
@@ -65,10 +65,10 @@ func TestClient_SetRootCertificateWatcher(t *testing.T) {
 		// (otherwise, test may succeed because 1st TLS session is re-used)
 		DisableKeepAlives: true,
 	}).SetRootCertificatesWatcher(
-		&CertWatcherOptions{PoolInterval: poolingInterval},
+		&CertWatcherOptions{PollInterval: poolingInterval},
 		paths.RootCACert,
 	).SetClientRootCertificatesWatcher(
-		&CertWatcherOptions{PoolInterval: poolingInterval},
+		&CertWatcherOptions{PollInterval: poolingInterval},
 		paths.RootCACert,
 	).SetDebug(false)
 
--- a/circuit_breaker.go
+++ b/circuit_breaker.go
@@ -48,23 +48,35 @@ type (
 		ApplyPolicies(*Response)
 	}
 
-	// CircuitBreakerObserver is an interface for observing circuit breaker events via hooks.
-	// It provides methods to register hooks for trigger and state change events, and to
-	// execute those hooks.
+	// circuitBreakerHookRunner is the dispatch side of [CircuitBreakerObserver].
+	// It stays unexported so a third-party CircuitBreaker is not required to
+	// implement Resty's internal hook plumbing to satisfy the public interface.
+	circuitBreakerHookRunner interface {
+		runOnTriggerHooks(*Request, error)
+		runOnStateChangeHooks(oldState, newState CircuitBreakerState)
+	}
+
+	// CircuitBreakerObserver is a [CircuitBreaker] that also reports its events
+	// via hooks. It embeds [CircuitBreaker] so a breaker with hooks attached can
+	// be passed straight to [Client.SetCircuitBreaker]:
+	//
+	//	client.SetCircuitBreaker(
+	//		resty.NewCircuitBreakerCount(3, 1, time.Second).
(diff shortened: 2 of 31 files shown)


jeevatkm (changes requested):
@mrueg Thanks for the PR. Can you please cover patch coverage? See codecov report.

mrueg:
@jeevatkm codecoverage should be fixed now

PR #1202, opened 2026-08-10
fix: stop handing out pooled buffers and aliased slices

mrueg:
Three memory-ownership defects, all the same shape: Resty handed out, or kept pointing at, memory it no longer owned.

## 1. The pooled request buffer stayed reachable through the `Request`

`Request.Execute` ended with `backToBufPool(r.bodyBuf)`, returning the buffer to `bufPool` without resetting it or clearing `r.bodyBuf`. Any later use of the same `Request` — `Clone`, `CurlCmd`, `fmtBodyString` — read a buffer another goroutine may already have taken from the pool. It is now reset and the field cleared; `backToBufPool` had no other caller and is gone.

## 2. Client-level values were aliased into every request

`parseRequestURL`, `parseRequestHeader` and `handleFormData` assigned client values into the request with `v[:]`, so the request's slice shared the client's backing array. When that array had spare capacity, a request's own `Add` wrote into it — and the next request, aliasing the same array, overwrote the value:

```
client:  tag=[a b c]           (len 3, cap 4)
req1:    tag.Add("first")   -> writes index 3
req2:    tag.Add("second")  -> writes index 3, same array
req1 now reads [a b c second]
```

Reachable through the public API: a client-level multi-value header or query param built with `Set` followed by two `Add`s leaves len 3 in a cap 4 array. Now cloned.

## 3. `Request.Execute` aliased its own retry slices

```go
retryConditions := append(r.retryConditions, r.client.retryConditions...)
```

appends into `r.retryConditions`' spare capacity, which `Request.Clone` shares with every clone, so executing one request could overwrite a condition registered on a clone. Now `slices.Concat`, reading the client lists through their locked accessors rather than the fields directly.

## Test change

`TestRequestClone` asserted that `parent.bodyBuf` was still readable after `Execute` — that is the behaviour fixed in (1). It now asserts the buffer was released.

## Verification

Four tests covering the three cases, all failing on `v3`. `go test ./... -race -count=1 -shuffle=on` green; `gofmt`, `go vet`, `staticcheck` clean.

Independent of #1201, #1202 and #1203; second wave from the same audit as #1197–#1200.

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -18,6 +18,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"slices"
 	"strings"
 )
 
@@ -156,7 +157,7 @@ func parseRequestURL(c *Client, r *Request) error {
 			if _, ok := r.QueryParams[k]; ok {
 				continue
 			}
-			r.QueryParams[k] = v[:]
+			r.QueryParams[k] = slices.Clone(v)
 		}
 
 		// GitHub #123 Preserve query string order partially.
@@ -189,7 +190,7 @@ func parseRequestHeader(c *Client, r *Request) {
 		if _, ok := r.Header[k]; ok {
 			continue
 		}
-		r.Header[k] = v[:]
+		r.Header[k] = slices.Clone(v)
 	}
 
 	if !r.isHeaderExists(hdrUserAgentKey) {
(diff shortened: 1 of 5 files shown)


jeevatkm (approved):
@mrueg Thanks for the PR.

PR #1203, opened 2026-08-10
fix: digest challenge parsing per RFC 7616

mrueg:
Two conformance problems in the `WWW-Authenticate` parser, both of which make Resty reject challenges real servers send.

## 1. qop tokens were not trimmed

```go
dc.qop = strings.Split(v, ",")
```

`qop="auth, auth-int"` parsed as `["auth", " auth-int"]`, and the leading space meant `isQopSupported` never matched the second token. So a server that offers a qop Resty doesn't know *first* — `qop="some-future-qop, auth"` — failed the request outright:

```
before: Get "http://…/dir/index.html": resty: digest: qop is not supported
after:  200 OK
```

even though `auth` was on offer. Each token is now trimmed.

## 2. Unknown directives were rejected

`setValue` returned `ErrDigestBadChallenge` from its `default` branch, so a single unrecognized auth-param anywhere in the challenge failed the whole request. RFC 7616 §3.3 requires unrecognized directives to be ignored. They are now skipped.

That change alone would have made an all-unknown challenge such as `Digest unknown_param=true` parse "successfully" into an empty challenge and then compute a credential from an empty realm and nonce. `nonce` is REQUIRED by §3.3, so `parseChallenge` now rejects a challenge carrying none — which is also what keeps the `/unknown_param` case in `TestClientDigestAuthErrors` failing for a real reason instead of passing silently.

## Verification

Added unit coverage for trimmed qop lists, ignored extension directives and the nonce requirement, plus an end-to-end test against the digest test server offering `qop="some-future-qop, auth"`. All fail on `v3`. `go test ./... -race -count=1 -shuffle=on` green; `gofmt`, `go vet`, `staticcheck` clean.

Independent of #1201, #1202 and #1204; second wave from the same audit as #1197–#1200.

### DIFF
--- a/digest.go
+++ b/digest.go
@@ -169,6 +169,13 @@ func (dt *digestTransport) parseChallenge(input string) (*digestChallenge, error
 		}
 	}
 
+	// nonce is REQUIRED by RFC 7616 section 3.3. Unrecognized directives are
+	// ignored, so this is what distinguishes a usable challenge from one that
+	// carried nothing we understand.
+	if isStringEmpty(c.nonce) {
+		return nil, ErrDigestBadChallenge
+	}
+
 	return c, nil
 }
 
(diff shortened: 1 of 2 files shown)


jeevatkm (approved):
Thanks @mrueg for the PR.

PR #1204, opened 2026-08-10
fix(enhancement): guard the debug-log type assertion and keep the rate limiter's context cause

mrueg:
Two small robustness fixes in unrelated files.

## 1. `debugLogger` asserted an interface value unconditionally

```go
Request: req.values[debugRequestLogKey].(*DebugLogRequest),
```

`prepareRequestDebugInfo` populates that key before the request is sent, so it is normally present — but when it is not, this panics inside a logging path:

```
panic: interface conversion: interface {} is nil, not *resty.DebugLogRequest
```

Now guarded with the two-value form and an empty fallback, so debug logging degrades instead of taking the process down.

## 2. Rate limiter errors erased the context cause

`RateLimitTokenBucket.Allow` and `RateLimitSlidingWindow.Allow` both returned the bare `ErrRateLimitExceeded` sentinel whether the wait ended in cancellation or a deadline, so callers could not tell the two apart and `errors.Is(err, context.Canceled)` never matched. The context error is now wrapped alongside the sentinel:

```go
errors.Is(err, resty.ErrRateLimitExceeded)   // still true
errors.Is(err, context.DeadlineExceeded)     // now also true
```

⚠️ **Behaviour change:** anyone comparing with `err == resty.ErrRateLimitExceeded` is affected; `errors.Is` is not. Happy to drop this half if you'd rather it waited for a major version.

## Verification

Both cases covered by added tests — the debug one panics on `v3`, the limiter one fails both `errors.Is` assertions. `go test ./... -race -count=1 -shuffle=on` green; `gofmt`, `go vet`, `staticcheck` clean.

Independent of #1201, #1202 and #1203; second wave from the same audit as #1197–#1200.

### DIFF
--- a/debug.go
+++ b/debug.go
@@ -137,8 +137,15 @@ func debugLogger(c *Client, res *Response) {
 		Body:       res.fmtBodyString(res.Request.DebugBodyLimit),
 	}
 
+	// prepareRequestDebugInfo populates this before the request is sent. Guard the
+	// assertion anyway so a logging path can never panic.
+	rql, _ := req.values[debugRequestLogKey].(*DebugLogRequest)
+	if rql == nil {
+		rql = &DebugLogRequest{}
+	}
+
 	dl := &DebugLog{
-		Request:  req.values[debugRequestLogKey].(*DebugLogRequest),
+		Request:  rql,
 		Response: rdl,
 	}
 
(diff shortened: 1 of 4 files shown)


jeevatkm (changes requested):
@mrueg Thanks for the PR. Can you please cover the patch test coverage? Please check the codecov report.

jeevatkm (approved):

PR #1212, opened 2026-09-06
api: fix the circuit breaker hook API and settle naming before the v3 freeze

mrueg:
> **Depends on #1208, #1209, #1210 and #1211.** Stacked on them; until they merge this PR shows their commits too.
>
> **Also conflicts with #1200**, which edits `AddRequestMiddleware` — renamed here. That PR's `max()` guard against a negative insert index is carried into `AddRequestMiddlewares`, so the fix is preserved either way; whichever lands second needs a rebase.

Every change here is breaking and becomes permanent at v3.0.0 final, so it belongs in an RC rather than after one. **This is the most opinionated PR of the five — please push back on anything you disagree with, or take it partially.**

## 1. `CircuitBreakerObserver` did not compile in its documented form

`OnTrigger` and `OnStateChange` returned `CircuitBreakerObserver`, which did not embed `CircuitBreaker`, so the obvious idiom was a hard build error:

```go
cb := resty.NewCircuitBreakerCount(3, 1, time.Second).
        OnTrigger(func(r *resty.Request, err error) {})
c.SetCircuitBreaker(cb)
```
```
cannot use cb (variable of interface type resty.CircuitBreakerObserver)
as resty.CircuitBreaker value in argument to c.SetCircuitBreaker:
resty.CircuitBreakerObserver does not implement resty.CircuitBreaker
(missing method Allow)
```

`CircuitBreakerObserver` now embeds `CircuitBreaker`, so a breaker with hooks attached goes straight into `SetCircuitBreaker`. The chaining methods moved from `circuitBreakerBase` — which is not a breaker on its own — onto `CircuitBreakerCount` and `CircuitBreakerRatio`.

`RunOnTriggerHooks` and `RunOnStateChangeHooks` are internal dispatch plumbing that any third-party `CircuitBreaker` had to implement to satisfy the public interface; they are now unexported. Added `State() CircuitBreakerState`, which had no exported accessor at all.

## 2. `RateLimiter.Allow` blocks, which is the opposite of what `Allow` means elsewhere

`golang.org/x/time/rate` — the limiter most people will adapt — defines `Allow() bool` as the **non-blocking** check and `Wait(ctx) error` as the blocking one. Resty's `Allow(ctx) error` blocks. This is the "same name, different semantics" trap: an adapter author reads `Allow`, wires it to `rate.Limiter.Allow()`, and silently gets a hard-reject limiter instead of a blocking one.

Renamed to `Wait` on the interface and both implementations.

## 3. `RetryDelayStrategyFunc` could not see the request or the attempt

The strategy received only `(*Response, error)`, so it had to reach through `res.Request` to learn which attempt it was on — and could not do even that when there is no response, which is the shape the SSE reconnect path passes. New signature:

```go
type RetryDelayStrategyFunc func(req *Request, res *Response, attempt int, err error) (time.Duration, error)
```

A correction worth recording: the audit that prompted this claimed the strategy was skipped entirely on transport errors. **That was wrong** — `Request.Execute` passes a non-nil `Response` wrapper with a nil `RawResponse`, so the old `res == nil` guard never triggered there. The signature change stands on its own merits, not on that rationale.

## 4. Naming

| current | becomes | why |
|---|---|---|
| `ContentDecompresser` (+5 more) | `ContentDecompressor` | not a word |
| `ErrNotHttpTransportType` | `ErrNotHTTPTransportType` | initialism |
| `CertWatcherOptions.PoolInterval` | `PollInterval` | it polls |
| `SRVWeightedRoundRobin.HttpScheme` | `HTTPScheme` | initialism |
| `HostStateInActive` | `HostStateInactive` | one word |
| `SSESource.SetSizeMaxBuffer` | `SetMaxBufferSize` | word order |
| `ResponseBodyUnlimitedReads` | `IsResponseBodyUnlimitedReads` | only bool getter without `Is`; nine siblings have it |
| `SetLoggerWarnLevel` | `SetDisableWarn` | `SetLoggerWarnLevel(true)` *suppressed* warnings while the getter is `IsDisableWarn` |
| `AddRequestMiddleware` | `AddRequestMiddlewares` (variadic) | only non-variadic `Add*`; arity cannot change later |
| `AddResponseMiddleware` | `AddResponseMiddlewares` (variadic) | same |
| `Client.ResultError` | unexported `resultErrorType` | returned `reflect.Type`, freezing the storage choice into the API, and collided by name with `Response.ResultError() any`; only `newErrorInterface` used it |

## 5. `Request.WithContext` shared mutable state with its source

```go
rr := new(Request)
*rr = *r
rr.ctx = ctx
```

`Header`, `QueryParams`, `FormData`, `PathParams`, `multipartFields` and `bodyBuf` were all shared, so `req.WithContext(ctx).SetHeader("X", "1")` mutated `req`. It now delegates to `Clone`, which already did this correctly.

## Verification

`go test ./... -race -count=1 -shuffle=on` green. `gofmt`, `go vet` clean. Coverage 99.9260%, up from 99.9233% on `v3`; the three remaining uncovered blocks are the same ones present before this work.

Last of five stacked PRs from an audit of the v3 branch.

### DIFF
--- a/cert_watcher_test.go
+++ b/cert_watcher_test.go
@@ -65,10 +65,10 @@ func TestClient_SetRootCertificateWatcher(t *testing.T) {
 		// (otherwise, test may succeed because 1st TLS session is re-used)
 		DisableKeepAlives: true,
 	}).SetRootCertificatesWatcher(
-		&CertWatcherOptions{PoolInterval: poolingInterval},
+		&CertWatcherOptions{PollInterval: poolingInterval},
 		paths.RootCACert,
 	).SetClientRootCertificatesWatcher(
-		&CertWatcherOptions{PoolInterval: poolingInterval},
+		&CertWatcherOptions{PollInterval: poolingInterval},
 		paths.RootCACert,
 	).SetDebug(false)
 
--- a/circuit_breaker.go
+++ b/circuit_breaker.go
@@ -48,23 +48,35 @@ type (
 		ApplyPolicies(*Response)
 	}
 
-	// CircuitBreakerObserver is an interface for observing circuit breaker events via hooks.
-	// It provides methods to register hooks for trigger and state change events, and to
-	// execute those hooks.
+	// circuitBreakerHookRunner is the dispatch side of [CircuitBreakerObserver].
+	// It stays unexported so a third-party CircuitBreaker is not required to
+	// implement Resty's internal hook plumbing to satisfy the public interface.
+	circuitBreakerHookRunner interface {
+		runOnTriggerHooks(*Request, error)
+		runOnStateChangeHooks(oldState, newState CircuitBreakerState)
+	}
+
+	// CircuitBreakerObserver is a [CircuitBreaker] that also reports its events
+	// via hooks. It embeds [CircuitBreaker] so a breaker with hooks attached can
+	// be passed straight to [Client.SetCircuitBreaker]:
+	//
+	//	client.SetCircuitBreaker(
+	//		resty.NewCircuitBreakerCount(3, 1, time.Second).
+	//			OnTrigger(func(req *resty.Request, err error) { ... }),
+	//	)
 	CircuitBreakerObserver interface {
+		CircuitBreaker
+
 		// OnTrigger registers one or more [CircuitBreakerTriggerHook] functions that are invoked
 		// each time the circuit breaker rejects a request in the open state.
 		OnTrigger(...CircuitBreakerTriggerHook) CircuitBreakerObserver
-
-		// RunOnTriggerHooks executes all registered trigger hooks with the given request and error.
-		RunOnTriggerHooks(*Request, error)
 
 		// OnStateChange registers one or more [CircuitBreakerStateChangeHook] functions that are
 		// invoked whenever the circuit breaker transitions between states.
 		OnStateChange(...CircuitBreakerStateChangeHook) CircuitBreakerObserver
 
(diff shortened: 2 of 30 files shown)


mrueg:
Force-pushed a correction, and #1200 / #1201 are now rebased on top of this branch so they no longer conflict.

**The correction:** the `ContentDecompresser` → `ContentDecompressor` rename listed in the table was not actually in the first push. I applied it with a word-boundary pattern (`\bDecompresser\b`), but that token only ever appears *inside* a larger identifier — `ContentDecompresser`, `AddContentDecompresser`, `contentDecompressers` — so the left boundary never held and the substitution matched nothing. Everything still compiled and every test passed, because the tree stayed internally consistent with the old spelling; the only visible effect was the error-message change I had made separately with a literal replacement, which left `ErrContentDecompresserNotFound` returning `"resty: content decompressor not found"`.

It is applied properly now:

```
ContentDecompresser            -> ContentDecompressor
AddContentDecompresser         -> AddContentDecompressor
ContentDecompressers           -> ContentDecompressors
ContentDecompresserKeys        -> ContentDecompressorKeys
SetContentDecompresserKeys     -> SetContentDecompressorKeys
ErrContentDecompresserNotFound -> ErrContentDecompressorNotFound
```

plus the unexported `contentDecompressers` / `contentDecompresserKeys` fields and `wrapContentDecompresser`. `grep -rn "ecompresser" --include='*.go' .` is now empty.

`go test ./... -race -count=1 -shuffle=on` green after the correction; `gofmt`, `go vet` clean.

PR #1214, opened 2026-09-08
fix: report load balancer outcomes before retries

### LINKED ISSUE #1213: Bug: LoadBalancerFeedback on connection io timeout
Hello, thank you for the project.

I’m trying to use the Load Balancer and found an interesting issue.

When a host is unreachable or the port returns a tcp reset (when the host is no longer listed), the Load Balancer doesn’t seem to receive any feedback about it. As a result, the next request may try to use the same host again. After a few attempts (`retryCount` greater than 1), it eventually reaches another healthy host.

What do you think about calling the feedback function when a network/tcp error occurs, so the Load Balancer can immediately mark that host as unhealthy?

add lines after  1520 line:

```go
			if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
				r.sendLoadBalancerFeedback(nil, err)
			}
```

https://github.com/go-resty/resty/blob/51f294a71849c21c3c4406b45de655a2c68588e0/request.go#L1515-L1530

and fix line

```go
success = !(errors.Is(noe.Err, syscall.ECONNREFUSED) || noe.Timeout())
```

https://github.com/go-resty/resty/blob/51f294a71849c21c3c4406b45de655a2c68588e0/request.go#L1866-L1871

Thanks.

fzlzjerry:
Fixes #1213.

A `net.OpError` timeout was marked successful by `!connectionRefused || timeout`. Feedback was also sent only after the retry loop, so a later successful attempt hid the earlier failed host from the load balancer.

Report each HTTP attempt before deciding on the next retry, and classify connection timeouts as failures. Preparation errors do not produce a backend report. A caller cancellation without an HTTP response is also excluded, while an actual HTTP response is still reported if cancellation or response processing fails afterward. The final attempt is not reported twice.

`RequestFeedback` now documents the per-attempt behavior. This leaves Windows connection-refused errno handling in #1192 separate.

Validation:
- 14 focused tests cover timeout classification, feedback before retry, weighted failover, exhausted/non-idempotent requests, HTTP status handling, connection refusal, caller cancellation/deadlines, preparation errors, and late cancellation after an HTTP response.
- The final tests reproduce the defects on unchanged `51f294a`; the HTTP status, connection-refused, and completed-response controls also pass on that baseline.
- Full suites pass on Go 1.23.12 and 1.27.0 with `-race -count=1 -covermode=atomic -coverpkg=./... -shuffle=on`.
- Focused tests pass with `-race -count=5 -shuffle=on` on Go 1.27.0.
- `go vet ./...` passes on both versions, and formatting/diff checks pass.
- All 3 added executable lines are covered by the Go coverage profile; overall statement coverage is 99.9%.

The transport failures in the tests are deterministic (`os.ErrDeadlineExceeded` inside `net.OpError`); the weighted failover test uses the real `WeightedRoundRobin` implementation.


load_balancer_timeout_test.go
```diff
@@ -0,0 +1,377 @@
+package resty
```
jeevatkm: @fzlzjerry Can you please move these tests into `load_balancer_test.go` file?

jeevatkm (changes requested):
@fzlzjerry Thanks for the PR. Please check the comment.

load_balancer_timeout_test.go
fzlzjerry: Moved all 14 feedback regression tests and their helpers into `load_balancer_test.go` in `c6cf79a00` and removed the separate test file. The declarations are unchanged; only the imports and file placement were consolidated.

Full race/coverage tests and `go vet ./...` pass on Go 1.23.12 and 1.27.1: 843 passing test/subtest observations and 99.9% statement coverage on each version.

Issue #630, opened 2023-03-14
DATA RACE: between go-resty/resty/v2.releaseBuffer()  and h2_bundle body 


zhongdalu-trechina:
I encountered a data race issue. Although the probability of it happening is very low, it still confuses me..

Operating environment is as follows.
```
golang version 1.18.10 

resty/v2 version 2.7.0

OS: win10
```

here is my test code:
```
var client = resty.New()

func main() {
	wg := sync.WaitGroup{}
	wg.Add(4000)
	for i := 0; i < 4000; i++ {
		index := i
		go func() {
			defer wg.Done()
			var data = make(map[string]interface{})
			data["index"] = index
			_, err := client.R().SetContext(context.TODO()).SetBody(data).Post("https://www.abc.com")
			if err != nil {
				return
			}
		}()
	}
	wg.Wait()
	fmt.Println("over")
}
```
here is DATA RACE warning:
1st:
```
==================
==================
WARNING: DATA RACE
Write at 0x00c0009a70f8 by goroutine 1418:
  bytes.(*Buffer).Reset()
      C:/Users/10188142.TREYANTAI/go/go1.18.10/src/bytes/buffer.go:99 +0x5a
  github.com/go-resty/resty/v2.releaseBuffer()
      D:/go/pkg/mod/github.com/go-resty/resty/v2@v2.7.0/util.go:287 +0x31
  github.com/go-resty/resty/v2.(*requestBodyReleaser).Close.func1()
      D:/go/pkg/mod/github.com/go-resty/resty/v2@v2.7.0/util.go:314 +0x44
  sync.(*Once).doSlow()
      C:/Users/10188142.TREYANTAI/go/go1.18.10/src/sync/once.go:68 +0x101
  sync.(*Once).Do()
      C:/Users/10188142.TREYANTAI/go/go1.18.10/src/sync/once.go:59 +0x46
  github.com/go-resty/resty/v2.(*requestBodyReleaser).Close()
      D:/go/pkg/mod/github.com/go-resty/resty/v2@v2.7.0/util.go:313 +0x7c
  net/http.(*readTrackingBody).Close()
      C:/Users/10188142.TREYANTAI/go/go1.18.10/src/net/http/transport.go:642 +0x56
  net/http.(*http2clientStream).abortStreamLocked()
      C:/Users/10188142.TREYANTAI/go/go1.18.10/src/net/http/h2_bundle.go:7097 +0xef
  net/http.(*http2clientStream).abortStream()
[...]

PR #1215, opened 2026-09-17
fix: potential data race issues

xxtanisxx:
I ran into a DATA RACE in Resty v3.0.0-rc.4, similar to https://github.com/go-resty/resty/issues/630. The added test reproduces the problem and passes with this fix. The fix is similar to how it was fixed in v2, not sure if there is better solution.

`go test -race -count=100 -run '^TestRequestBodyPoolLifetime$' .` 

Without the fix:
Screenshot 2026-09-16 at 8 04 15 PM

With fix:
Screenshot 2026-09-16 at 8 04 42 PM


jeevatkm:
@xxtanisxx Thanks for creating a PR. I believe I addressed the data race in the v3. I will double-check and get back to you.

Voyagerroc-Lab (approved):
Reviewed with assistance from OpenAI Codex. Cloning the pooled buffer into an independent reader prevents the request body from observing storage after the pool reuses it. The regression test reads the body after the request returns, which exercises the lifetime issue directly.

xxtanisxx:
@jeevatkm Thank you so much!!!

PR #1218, opened 2026-09-19
fix: handle error from cookiejar.New in createCookieJar

clydeeshun94:
cookiejar.New may return an error if the PublicSuffixList fails to load. Silently discarding this error could result in a nil *cookiejar.Jar being returned, causing panics on subsequent cookie jar operations.

Fall back to cookiejar.New(nil) which returns an empty jar on error.


jeevatkm:
@clydeeshun94 Can you please add test case for the PR?

jeevatkm:
@clydeeshun94 Thanks for adding the test case. It seems the CI build failed with the test case. Can you please check it?

PR #1219, opened 2026-09-19
fix: handle error from cookiejar.New in createCookieJar

clydeeshun94:
## Summary

cookiejar.New may return an error if the PublicSuffixList fails to load. The original code silently discarded this error, which could result in a nil *cookiejar.Jar being returned and causing panics on subsequent cookie jar operations.

## Fix

Fall back to cookiejar.New(nil) (returns an empty jar) when the primary call fails.

## Changes

- **resty.go**: Check error from cookiejar.New, fall back to empty jar on failure
- **client_test.go**: Add TestCreateCookieJar to validate the fix


jeevatkm:
@clydeeshun94 You have targeted the incorrect branch. Plus, instead of updating the existing PR #1218, you have created a new one.

clydeeshun94:
Crap. This is a duplicate of #1218. PR #1218 correctly targets the v3 branch and now includes the fix and the test case. Closing this.

PR #1220, opened 2026-09-20
feat(enhancement): accept all HTTP-date formats in Retry-After

agammann:
Retry-After dates are parsed only as RFC 1123. Valid RFC 850 and asctime dates are ignored, so retries fall back to the normal delay instead of honoring the header.

Use net/http.ParseTime, which accepts all three HTTP-date formats required of recipients by RFC 9110 section 5.6.7: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7. Extend the existing test table with future and past dates in both legacy formats.

Validation with Go 1.27.1: all four new cases fail before the fix. The full go test ./... suite passes on Linux after it. The Windows full-suite run encounters separate existing load-balancer, output-path, and trace expectations; the passing full-suite result is the Linux run.


jeevatkm (approved):
@agammann Thanks for the PR.

